PR #4, opened 2016-04-18
Create backup of native Smart House demo code.

uhofemei:


gvancuts:
I don't understand what the purpose of this PR is, can you explain please?

uhofemei:
I created a backup branch containing the native code OCF demo bits. This PR populates is. I didn't remove the bits from the master branch yet for safe measure until we could talk about it first.

gvancuts:
I'm not sure we need this. All the code is and will remain part of the history. If we want it to be more easily accessible then a branch makes sense but I would give it a name that is more explicit about its content.

uhofemei:
I'm not sure we need the native code bits at all to be honest. The code is based on a pre 0.9 version of Iotivity and the world has changed significantly since then. Can we decide to only feature the Iotivity 1.0.1 based high-level JS API version of the demo?

gvancuts:
I've just tagged the latest version in the Git history that has all the bits and pieces of the code. I agree I don't think we need to do anything else with it anymore. My suggestion would be to remove all four folders associated with the code and make the content of the dollhouse folder the main content.

poussa:
@gvancuts +1

Lets do it!

gvancuts:
Closing this as per the previous comments.

PR #11, opened 2016-05-24
Implement sensor servers using Soletta I/O Node.js bindings

nagineni:
This branch includes the implementation of GPIO and AIO based sensor servers using Soletta I/O Node.js bindings.

### DIFF
--- a/ocf-servers/soletta-js-servers/README.md
+++ b/ocf-servers/soletta-js-servers/README.md
@@ -1,0 +1,49 @@
+# JavaScript OCF servers
+This folder contains JavaScript implementation of various OCF servers, such as:
+* Fan
+* CO2 detector
+* Motion Sensor
+* RGB LED
+* LED (on/off)
+* Buzzer
+* Ambient Light Sensor
+* Temperature
+* Solar Panel
+* Button
+* Switch
+
+# How to start these OCF servers
+## Prerequisites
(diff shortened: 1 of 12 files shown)


gvancuts:
I have not tested this but this looks good to me.

Just to make sure I understand it correctly, it does not (yet?) use Soletta OCF's implementation?

nagineni:
@gvancuts Thanks for looking into this.

> I have not tested this but this looks good to me.

I've been testing these sensors with the demo setup we have here and the outcome is exactly as same as the sensors based on mraa.

> Just to make sure I understand it correctly, it does not (yet?) use Soletta OCF's implementation?

You are right. These are still based on the iotivity-node. I will update to Soletta OCF implementation, once the bindings are ready.

nagineni:
@poussa

nagineni:
@gvancuts Could you please merge this?

Thanks!

gvancuts:
@poussa , OK to merge?

poussa:
LGTM. I like the idea we keep both versions in separate directories

js-servers: mraa and iotivity-node
soletta-js-servers: soletta HW and soletta OCF (once ready)

gvancuts:
I fully concur, merging.

/me also making a mental note that it'll soon be time to update the top-level README.md to decribed this new folder (and also the ostro-config/systemd-services folder which we do not talk about).

PR #12, opened 2016-05-26
Update OCF servers implementation to the latest JS spec

nagineni:
This removes suffix "Resource" from (un)?registerResource as per the spec.


nagineni:
@poussa @gvancuts

nagineni:
@gvancuts Could you please land this?

PR #13, opened 2016-06-14
Use debuglog to enable logging conditionally

nagineni:
This will disable the console logging by default, but users can enable
it by running the program with 'NODE_DEBUG=<module name>'.

Client: NODE_DEBUG=first_server node first_server.js
Sensor: NODE_DEBUG=fan node fan.js


gvancuts:
:+1:

nagineni:
@poussa

poussa:
Sweet ! I like it.

PR #14, opened 2016-06-15
 Call subscribe() method to enable global presence flag

nagineni:
Call subscribe() method to enable global presence, so that the client will receive device presence notifications "added", "deleted", "changed"


nagineni:
@poussa @gvancuts

gvancuts:
LGTM but I haven't tested this

nagineni:
Tested this with the latest iotivity-node and it working fine.

poussa:
LGTM !

PR #20, opened 2016-07-05
Add support to control the solar panel using 3D UI or REST server

nagineni:
Add support to control the solar panel using the UI element in the 3D model UI or setting simulationMode property ON/OFF with the REST server. When the simulation mode is on, the panel stands up and it will slowly tilt back during the day and forward in the night. The time information also be updated on the LCD while the panel is moving back/forward.


nagineni:
@poussa @gvancuts @uhofemei Could you please take a look? This will allow us to control the solar panel without installing icarus and other broken python dependencies. Also, we can setl the panel ON/OFF using the solar panel UI element in the 3D model UI.

poussa:
LGTM. Did not test. This removes the major pain (python) from the demo.

PR #22, opened 2016-08-19
Update client and servers implementation to IoTivity 1.1.1

nagineni:
This updates the client and servers implementation to IoTivity 1.1.1.

### DIFF
--- a/ocf-servers/js-servers/ambient_light.js
+++ b/ocf-servers/js-servers/ambient_light.js
@@ -104,7 +104,9 @@ function retrieveHandler(request) {
 }
 
 device.device = Object.assign(device.device, {
-    name: 'Smart Home Illuminance Sensor'
+    name: 'Smart Home Illuminance Sensor',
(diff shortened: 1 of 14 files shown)


nagineni:
@poussa @gvancuts @uhofemei PTAL

poussa:
What happens if you dont set the coreSpecVersion and dataModels variables?

nagineni:
Unfortunately coreSpecVersion and dataModels are not an optional attributes and server(s) will throw TypeError while setting the device info.

@zolkis should we mark them as optional in OicDevice info?

zolkis:
> What happens if you dont set the coreSpecVersion and dataModels variables?

AFAIK a device always needs to set these, since it's needed for the OCF certification (i.e. against which spec version and data model it is certified).

poussa:
OK, they are mandatory and code throws exception if they are not set. That is all clear now.

+1

gvancuts:
:+1:

PR #24, opened 2016-09-13
Enable light weight rules engine for the Smart Home sensors 

nagineni:
This branch contains the changes needed for enabling the light weight rules engine for the Smart Home  gateway. The rules engine takes the rules written in JSON format as input. It evaluates payload data received from the sensors with each rule and triggers the registered action when the conditions are met.

The following rules are now included:
1. Change RGB LED color based on motion detection
2. Trigger fan ON/OFF based on the gas detection

I also added an option (-r) to run the service only with the rules engine just in case if we don't want to start 3D model UI.


nagineni:
@poussa @gvancuts @uhofemei PTAL

nagineni:
Tested this on the gateway and it's working fine.

first_server.js
```diff
@@ -1,64 +1,117 @@
...
 
-var connectionList = {};
+            connection.on('message', function(message) {
+                if (message.type === 'utf8') {
+                    debuglog('Received Message: ' + message.utf8Data);
 
-function originIsAllowed(origin) {
-    // put logic here to detect whether the specified origin is allowed.
-    return true;
+                    parceInComingRequest(JSON.parse(message.utf8Data), connection);
```
gvancuts: Nitpick: should we change this to parseInComingRequest (instead of 'parce...')?

first_server.js
```diff
@@ -1,64 +1,117 @@
...
+            });
+            connection.on('close', function(reasonCode, description) {
+                debuglog((new Date()) + ' Peer ' + connection.remoteAddress + ' disconnected.');
+                delete connectionList[connection.remoteAddress];
+            });
+        }
+    });
 }
 
 function parceInComingRequest(message, connection) {
```
gvancuts: Same as above

first_server.js
```diff
@@ -72,56 +125,29 @@ function updateWebClients(msg, eventType) {
...
-    for (var key in connectionList) {
-        debuglog(key);
-        var connection = connectionList[key];
-        connection.sendUTF(JSON.stringify(outmesg_list));
-        //connection.sendUTF();
+    var newEvents = rulesEngine.processEvents(JSON.stringify(outmesg_list));
+    if (newEvents) {
+        var actions = JSON.parse(newEvents);
+        for (var action in actions) {
+            parceInComingRequest(actions[action]);
```
gvancuts: Same as above

gvancuts:
This is great stuff - thanks @nagineni !

I have just one comment (a nitpick really) and it's about the name of a function you use internally.

first_server.js
nagineni: totally make sense :)

nagineni:
Fixed the comment and rebased on master.

gvancuts:
:+1:

gvancuts:
I'm tempted to actually make running the service as a rule engine the default (i.e. adding '-r' to the systemd service file). 

@nagineni @poussa thoughts?

nagineni:
@gvancuts I am fine with that.

uhofemei:
Works for me too. I wish it were easier to update the 3D model.

poussa:
LGTM. I support having this as default.

nagineni:
Thanks all! Updated the systemd service file to start the first_server in rule engine mode default.

PR #27, opened 2016-09-21
Fix bugs in temperature and sensor rules

nagineni:
This PR includes the following bug fixes.

- Handle undefined queryOptions in temperature retrieve handler.
- Update rules to set the RGB LED color to Blue by default


nagineni:
@uhofemei @gvancuts @poussa

uhofemei:
Tested and works. 👍

PR #32, opened 2016-09-30
Dockerfile: fix URL for cloning SmartHome-Demo repo

gvancuts:
The URL listed in the Dockerfile (smarthome-web-portal/tools/docker)
leads to an error when attempting to build the container image. An
extra "https://" gets inserted and makes cloning the SmartHome-Demo
fail.

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

PR #33, opened 2016-09-30
Update the project folder in the conf file for the new repo.

ichbinblau:
The repo moved to 01org. So the process directory in the supervisor configuration file also needs change.

### DIFF
--- a/smarthome-web-portal/tools/docker/web-portal.conf
+++ b/smarthome-web-portal/tools/docker/web-portal.conf
@@ -7,7 +7,7 @@ autorestart=unexpected
 [program:admin-portal]
 priority=997
 command=/usr/bin/python2.7 run.py
-directory=/opt/smarthome-web-portal/tools/smarthome-admin-portal/
+directory=/opt/SmartHome-Demo/smarthome-web-portal/tools/smarthome-admin-portal/
 autostart=true
 autorestart=true
 redirect_stderr=true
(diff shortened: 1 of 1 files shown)


gvancuts:
This looks good to me @ichbinblau but I am wondering if these two lines shouldn't also be updated:
- https://github.com/01org/SmartHome-Demo/blob/master/smarthome-web-portal/tools/docker/Dockerfile#L65 (this is just a comment but it would make it cleaner
- https://github.com/01org/SmartHome-Demo/blob/master/smarthome-web-portal/tools/docker/Dockerfile#L85

Since I have already touched the Dockerfile, I can add this to my PR (#32) instead (but please confirm whether it's the right course of action).

ichbinblau:
@gvancuts 

- https://github.com/01org/SmartHome-Demo/blob/master/smarthome-web-portal/tools/docker/Dockerfile#L65   (You can either remove this line or update the path as this config file is no longer needed. )

- https://github.com/01org/SmartHome-Demo/blob/master/smarthome-web-portal/tools/docker/Dockerfile#L85 ( Yes, you are right. Please change it to `ENV PYTHONPATH $PYTHONPATH:/opt/SmartHome-Demo/smarthome-web-portal`) 

And I just did a complete test on the change. It works with the new repo after these updates.

gvancuts:
Merging as I got a 👍  from @ichbinblau over email

PR #36, opened 2016-10-06
Create 'baseline' branch for validation

ttzeng:
Since the demo depends on the iotivity and iotivity-node included in the underlying Ostro image, please create a 'baseline' branch from the release '1.1.0-5' for incremental merging the later commits.


gvancuts:
This would merge your changes into master whereas I believe you intention is to create a branch where you can use 1.1.0-5 as the baseline code and cherry-pick from master to add new features without forcing the use of IoTivity 1.1.1. If my assumption is correct, I think I would need to manually create that branch for you so you can then create Pull Requests against it.

gvancuts:
I have created the [1.1.0-5-branch](https://github.com/01org/SmartHome-Demo/tree/1.1.0-5-branch) for  you. You should now be able to create Pull Requests for it.

ttzeng:
Thanks for creating the new branch for doing the cleaning job. But could
you use the different branch name? once we've successfully running the demo
against the new Ostro image with iotivity-node 1.1.1, the same branch
should still be used. The name '1.1.0-5-branch' will be confusing, IMO.

nagineni:
@gvancuts thanks for creating a new branch. We have a release tag for that branch already, so we can re-create it or make a new tag if we cherry-pick any changes from master to the  branch.

@ttzeng, lets have a separate branch for each iotivity release. Current master is based on v1.1.1 and I will create a new branch with 1.1.1-<release number>, before I start updating master to iotivity v1.2.0.

nagineni:
@ttzeng some of the changes in master needs iotivity v1.1.1, so check the changes if you are planning to cherry-pick commits from master to 1.1.0-5-branch.

gvancuts:
Hi @ttzeng , what I had in mind was what @nagineni describes so I'd rather stay with the name (and I agree we should also update the tag if and when we cherry-pick enhancements or add new ones).

Does that work for you @ttzeng ?

ttzeng:
@gvancuts <https://github.com/gvancuts>, I'd rather have a single branch
keeping the BKC of the demo, and label tags to identify milestones. Having
branches for each iotivity revision may result several zombie branches, as
once we have stable configuration for newer iotivity, there won't need
further efforts on the last BKC.

@nagineni <https://github.com/nagineni>, we could submit PRs only for
"branches", the tags don't work. So far we can only submit PRs for "master"
or "backup", that's why everybody send PRs to patch the master branch, but
we only validate against our local dependencies which causes the problem
IMHO.

ttzeng:
Or a branch name 'ostro' reflects this new branch is going to track the ostro builds. The plan is to cherry pick commits that have been validated on that new Ostro build (i.e. BKC, Best Known Configuration), so people who are interested to replicate the demo only need to checkout this particular (say 'ostro') branch, he/she doesn't care the iotivity revisions. Does it make sense to you?

nagineni:
>  @nagineni <https://github.com/nagineni>, we could submit PRs only for "branches", the tags don't work.

I knew  :smiley: .  Earlier there was no plan in backporting the commits to the demo based on old version(s) of iotivity, and that was the reason for having only release tags instead of branches. I happy to see the interest in back-porting the changes now.

nagineni:
> Or a branch name 'ostro' reflects this new branch is going to track the ostro builds. The plan is to cherry pick commits that have been validated on that new Ostro build (i.e. BKC, Best Known Configuration), so people who are interested to replicate the demo only need to checkout this particular (say 'ostro') branch, he/she doesn't care the iotivity revisions. Does it make sense to you?

What happens if someone takes the old Ostro build that contains some old version of iotivity and other relevant components?

gvancuts:
In my opinion, it's best to keep branches based on IoTivity version as we are trying to decouple the code from any specific OS (so it could run on Ubuntu, Fedora, Yocto, etc.). I also think that within a branch, we won't require an upgrade of IoTivity. We will have 'zombie' branches as we create newer ones (that are at that point, by definition, stable and considered a BKC) because there will be less interest in running older code. But having it available and known to work would still be good.

The way I see this going, master is where we bring all new code, new versions, enhancements, etc. 'master' could break at times (although we should try to avoid that in general). Some of these enhancements would be cherry-picked and applied to the BKC branches based on our needs. We could also introduce specific bug fixes in those branches (that may not be needed on master) but the point would be that whatever we do on those branches should have been well tested before merging.

Does that make sense?

nagineni:
:+1:

ttzeng:
In case someone takes an old Ostro build, he can clone the 'ostro' branch
which contains all code that have been validated on Ostro builds, then
checkout the tag specific to that build (assuming that build was BKC ago).
As long as the demo code has been validated on a particular Ostro build,
there is a tag associated with that build. Whoever want to replicate the
demo on Ostro, only need to clone the 'ostro' branch, he may not know which
iotivity version included in that build.

On Fri, Oct 7, 2016 at 7:55 PM, Sudarsana Nagineni <notifications@github.com

ttzeng:
I am afraid decouple the demo code from OSes may create more uncertainty as
iotivity is officially supported on Tizen, Ubuntu, Android, Windows as
highlighted in iotivity.org, and are we sure Ubuntu supports well on all
our hardware platforms such as Edison?

nagineni:
> In case someone takes an old Ostro build, he can clone the 'ostro' branch which contains all code that have been validated on Ostro builds, then checkout the tag specific to that build (assuming that build was BKC ago). As long as the demo code has been validated on a particular Ostro build, there is a tag associated with that build. Whoever want to replicate the demo on Ostro, only need to clone the 'ostro' branch, he may not know which iotivity version included in that build.

Does this mean we will have a new release tag for each Ostro build? What about if someone wanted to try the demo on different platform(e.g linux pc) with some specific version of iotivity-node?

PR #40, opened 2016-10-13
Set correct type for the 'range' & 'rgbValue' properties of Color RGB resource

ttzeng:


nagineni:
LGTM. 

Nice that it has been changed from string to array in the latest data model http://oneiota.org/revisions/1414

gvancuts:
Thank you both. Does that require any changes anywhere else (e.g. web portal code)?

nagineni:
Yes, probably in the Cordova android app. No changes required in the web portal code.

gvancuts:
OK, so no further changes should be made on this repo to keep things working. I'll merge it. Thanks

PR #43, opened 2016-10-14
Add a Zephyr environmental sensing application running on Arduino 101

kiendinh:


kiendinh:
A new environmental sensing app has been added. It measures temperature, humidity, pressure and UV index and send data over BLE to connected device using READ/NOTIFY. @gvancuts, @nagineni, @ttzeng, @uhofemei please review. Thanks.

gvancuts:
Thanks for developing this @kiendinh 

What I am missing at the moment is how this integrates with the rest of the demo (and more specifically IoTivity). Can you clarify?

kiendinh:
@gvancuts Yes, that's what I'm working on now, trying to connect it to an Edison using noble. However, somehow there is some limitation of Bluetooth support in the Edison (due to kernel version), and noble as well, that the data still cannot be read. It is alright to use an Android phone to connect and receive data by reading/notifying.

gvancuts:
Have you tried to connect it to the Gateway instead (using a dongle)? We may be phasing out the Edison boards soon so we'll need another board to connect it to and the Home Gateway seems like a natural fit to me.

For my own education, could you explain how OCF/IoTivity is used in this case?

kiendinh:
It works on other devices like Joule, Minnow/Galileo + BLE dongle. The OCF/IoTivity server (under construction) will be on the devices like other js-servers.

sensors/environmental-sensors/README.md
```diff
@@ -0,0 +1,67 @@
...
+data from a set of sensors connected to the Arduino 101 and sends it to the SoC
+through IPM. The collected sensor data is also displayed by the sensor subsystem
+on a Grove LCD.
+
+The SoC application exposes the received sensor data as a simple Bluetooth
+Environmental Sensing Service.
+
+## Sensor wiring:
+
+The sample uses the BME280 sensor for temperature, humidity and air pressure
```
wanghongjuan: Hi, @kiendinh Could you please add the sensor link? Just as you mentioned that required HW in [DC_power_meter](https://github.com/01org/SmartHome-Demo/tree/master/sensors/DC_power_meter#required-hw)

gvancuts:
Thanks for the clarifications @kiendinh . Could you enhance the README.md file with such information? I'm thinking that a new chapter in it that explains how one can collect the data on the receiving end (ref app on Android or iOS, other node.js app on the Gateway using noble, etc.) would make it clear to everyone. We should also make it clear if it's still a little disconnected from the rest of the demo today because of the lack of OCF support but mark that as a future item that is being worked on.

sensors/environmental-sensors/README.md
kiendinh: I just added links to the sensors.

kiendinh:
I just added some more details to mention about the receiving end.

gvancuts:
Thanks @kiendinh , I will be merging this now.

In the future, can you use "git commit --amend" when making modifications to an existing commit instead of adding new ones? It's OK in this case as GitHub allows me to squash all commits into one when merging so I will do that.

PR #45, opened 2016-10-25
[ocf-servers] Add an environmental sensor OIC resource.

kiendinh:
It uses noble to connect via BLE to the environmental sensor to get data.
There are temperature, humidity, air pressure and UV index sensor data that
the resource represents. To use, launch the nodejs application in the
homegateway with BLE capability. It has been tested on MinnowBoard Turbot &
Galileo with a CSR 4.0 BLE USB dongle. Edison is not yet supported.

### DIFF
--- a/ocf-servers/doc/environment.raml
+++ b/ocf-servers/doc/environment.raml
@@ -1,0 +1,35 @@
+#%RAML 0.8
+title: OICEnvironmentalSensor
+version: v1.0-20150727
+schemas:
+  - Environment: !include oic.r.sensor.environment.json
+traits:
+  - interface:
+      queryParameters:
+        if:
+          enum: ["oic.if.s"]
+
+/EnvironmentSensorResURI:
+  description: |
+    This resource describes an environmental sensor
+    temperature (Celcius), humidity (percentage) and pressure (hPa) are float numbers,
+    while uvIndex is an integer. They represent the sensed environemental data.
+
+  displayName: Environtmental Sensor
+  is: [ interface ] # valid for all methods
+
+  get:
+    responses:
+      200:
+        body:
+          application/json:
+            schema: Environment
+            example: |
+              {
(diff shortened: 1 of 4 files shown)


gvancuts:
This looks good to me. Could someone independently verify this before we merge it?

I think we should also make a note that the ultimate goal would be to have iotivity-constrained running on the Zephyr device.

ttzeng:
We might create a custom resource type incorporating temperature, humidity, UV radiation properties, or use three standard resource types - oic.r.temperature, oic.r.humidity, oic.r.sensor.radiation.uv defined in OCF spec (http://oneiota.org/)

kiendinh:
I was looking for a way to include different resource types under one URI following the specs, but so far it didn't seem to work. When iotivity-node is ready, we can switch to using standard types fairly easy.

This is kind of a test of BLE working in our system. It's proved to work and we can gradually replace wire with wireless connection and offload sensor tasks to sensor devices.

ttzeng:
Though current iotivity-node does not handle the 'rt' property correctly, you might try to register the device with resourceType set to the array of resource types, and incorporate all the required values to a single property for a try.

ttzeng:
After verifying using Iotivity base API, the device registers with multiple resource types only can be discovered by using the 1st resource type, so we may use custom resource type or wait the fix in iotivity-node, IMHO

gvancuts:
it looks like this problem was fixed earlier, see https://github.com/otcshare/iotivity-node/commit/025cdc654c69c66aca69329e703ffa987523fafa

ttzeng:
That commit hasn't been merged in the master branch, we don't know whether it has been validated. Since the demo is depending on iotivity-node module, I give up the idea to use standard resource types for this server, suggest to use the custom resource type instead.

kiendinh:
I'd like to close this one and submit the new one which contains simulation mode for testing.

PR #52, opened 2016-11-09
Cloud portal code update before Intel 360. 

### LINKED ISSUE #68: zero value for solar/temperature/illuminance sensor data not updated
When parsing data after REST client getting the sensor data via REST API, there's something wrong with the condition to update data when the sensor data happen to be 0. For example, the [solar panel data parsing](https://github.com/01org/SmartHome-Demo/blob/master/smarthome-web-portal/CeleryTask/managers/data_manager.py#L101) is represented as:
```
                if sensor == 'solar':
                    new_tilt = data.get('properties').get('tiltPercentage')
                    if new_tilt:
                        content.update({
                            'tiltpercentage': new_tilt,
                            'lcd_first': data.get('properties').get('lcd1'),
                            'lcd_second': data.get('properties').get('lcd2'),
                        })
                        # add_method(content)
                        print 'update tilt percentage: {}'.format(str(add_method(content)))
                    else:
                        print "Unable to get tilt percentage."
```

Here, <code>new_tilt</code> may get a value of 0, which is the initial value when creating solar OCF resources. Unfortunately, in Python, both <code>bool(0)</code> and <code>bool(None)</code> return False. Thus, when the tiltPercentage is 0, content will not update. 
This may cause the following behavior:
- The browser may not display the resource even it exists. This happened in my tests when the solar panel resource was created. The log said: 
  > Unable to get tilt percentage.

  After changing the tiltPercentage to a non-zero value via an OCF client, the solar panel resource could show up in browser.

- In browser, the tiltPercentage of solar panel will never be 0. Not sure if this is reasonable.

This issue also exists in temperature and illuminance resources. I think for temperature data, 0 degree in Celsius is reasonable and should update.

ichbinblau:
1. Add the new sensor type and display them in Web UI
2. Moved all the global constants to settings.py
3. Tagging and grouping feature
4. Use 'di+path' for resource identification
5. Fixed #68, #74 
6. Add data analytics feature in the home dashboard and the admin portal
7. Support for Juke box
8. Portal updates for iotivity 1.2.x


gvancuts:
Sorry for the delay in reviewing this @ichbinblau . This looks good to me! Is there anyone who could test this independantly? @uhofemei @kiendinh @ttzeng ?

kiendinh:
Looks good to me, too.

gvancuts:
@kiendinh , does it mean you've tested it or done a visual inspection of the code? Tx!

kiendinh:
@gvancuts, I only reviewed the code.

smarthome-web-portal/CeleryTask/managers/data_manager.py
```diff
@@ -136,13 +136,25 @@ def parse_data(data, **kargs):
                     new_rgb = data.get('properties').get('rgbValue')
                     if new_rgb:
                         content.update({
-                            'rgbvalue': new_rgb,
-                            'range': data.get('properties').get('range'),
+                            'rgbvalue': str(new_rgb),
```
ttzeng: The 'rgbValue' and 'range' properties are now array of integers (by [this commit](2c4c397aadd5b2162a30f5d71a1dce7cc96c7574) per the [OCF spec](http://oneiota.org/revisions/1758)) .

ichbinblau: Yes, they are sets of rgb values. I converted them to String  and save them in database and parse them when visualizing the data.

smarthome-web-portal/CeleryTask/managers/data_manager.py
```diff
@@ -165,25 +177,26 @@ class DataManager(base.BaseTask):
...
+    def get_sensor_types_map():
+        types = sensor_type.get_all_types()
+        mapping = dict()
+        for typ_dict in types:
+            if typ_dict['type'] == 'motion':
+                pth = '/a/pir'
+            elif typ_dict['type'] == 'environment':
+                pth = '/a/env'
+            else:
+                pth = '/a/' + typ_dict['type']
```
ttzeng: Each server can register its own id, why assumes the id is tied to the type name?

ttzeng:
I cloned the target branch of the repo from Theresa's home, and modify the Dockerfile to pul the code from Theresa's target branch instead of the 01.org code. The image can be built, but it can't show my 3 pseudo LED servers. I also observed some errors in the docker log, below is one of the error:
[2016-11-23 10:05:30,449: WARNING/Worker-2] http://192.168.1.103:8000/api/oic/a/pir?di=31653931-6330-6339-2d33-3839322d3431
[2016-11-23 10:05:30,450: INFO/Worker-2] Starting new HTTP connection (1): 192.168.1.103
[2016-11-23 10:05:30,808: WARNING/Worker-2] Exception in thread 31653931-6330-6339-2d33-3839322d3431:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/opt/SmartHome-Demo/smarthome-web-portal/CeleryTask/managers/data_manager.py", line 51, in run
    self.client_conn.get_data(stream=True, callback=self.parse_data, **kargs)
  File "/opt/SmartHome-Demo/smarthome-web-portal/RestClient/sensor.py", line 53, in get_data
    self.resp.get_data(callback, **kargs)
  File "/opt/SmartHome-Demo/smarthome-web-portal/RestClient/api/__init__.py", line 200, in get_data
    parser(json.loads(chunk), **kargs)
  File "/opt/SmartHome-Demo/smarthome-web-portal/CeleryTask/managers/data_manager.py", line 93, in parse_data
    add_method = get_class("DB.api.{}.new".format(sensor))
  File "/opt/SmartHome-Demo/smarthome-web-portal/utils/util.py", line 202, in get_class
    m = __import__(module)
ImportError: No module named pir

ichbinblau:
Hi, @ttzeng. I will check the error on my side. BTW, if you cannot see the  pseudo LED, would you make sure that they are registered in the gateway and report data to the server? Sometimes I found that it may require a server restart if there are some new sensor coming in,

ichbinblau:
Hi, @ttzeng . Thx for the finding. I changed the pre-staged sensor types in my database but not the script. I have submitted a  commit to solve the "No module named pir" error. But I am able to see the simulated LED sensors in the web portal. Would you make sure that the sensors are connected with the gateway? 
capture

ttzeng:
@ichbinblau can you please rebase the code to at least [this commit](https://github.com/01org/SmartHome-Demo/tree/cd3f6cf052373b46f5f02505990eb8700e8d4de9) so that it   's self contained without pulling the code from github, and it will be easier for validation. Thanks in advance.

ichbinblau:
HI, @ttzeng I have updated to the upstream.

gvancuts:
Hi @ichbinblau , how did you update to include the latest commits? Did you cherry-pick and apply them to your branch?

I'm asking because this PR now includes all these commits. I would suggest to instead roll back to where you were and rebase, so something like:
$ git reset --hard 6795a1c2b22e9db50156f4cdfed217b2899a4d74
$ git remote add 01org https://github.com/01org/SmartHome-Demo
$ git fetch 01org
$ git rebase 01org/master
$ git push -f origin environmental_sensor

It's considered good practice to rebase your PR when you know master has evolved, we don't always do it and as long as there is no merge conflicts, GitHub does not enforce that but we should perhaps get in the habit of doing it regardless.

ichbinblau:
Hi, @gvancuts. I did not use cherry-pick. Here are the steps I followed .  I guess it was caused by the second rebase pull which overturned sequence.  

$ git remote add upstream[https://github.com/01org/SmartHome-Demo](https://github.com/01org/SmartHome-Demo)
$ git fetch upstream
$ git rebase upstream/master
$ git pull --rebase origin environmental_sensor # updated to the commit 6795a1c cos the commit was not up-to-date.
$ git push origin environmental_sensor

Thanks for showing me the correct steps. I think it should be right now.

ttzeng:
By validating the [ZIP file](https://github.com/01org/SmartHome-Demo/tree/7a0fb6b4aa39f92ea1a2e95dc15a0987baa88267) with the 3 commits of this PR:
(1) the docker image built successfully (i.e. the previous 'pir' failure has been resolved) -- OK
(2) the admin and smart home portal run -- OK
(3) the data of environmental sensors (simulation mode) can be shown on additional 4 cards in the portal -- OK
(4) run my 'pseudo' led server 4 times with different ID (led0, led1, led2, led3), no any led cards shown and errors in the docker log file [a.txt](https://github.com/01org/SmartHome-Demo/files/638632/a.txt) -- Fail

ichbinblau:
Hi, @ttzeng. Thanks for the report. I found that the sensor path in the log file was **'/a/led3'** instead of '/a/led' . Would you change it back to '/a/led' ? The uuid should remain the same after the change and restarting the ocf server. 
My code only allows registered sensor types and path to join in the Cloud.

ttzeng:
Then, what did you mean by "support of multiple sensors of the same type"? according to OCF 1.1 spec, section 7.3.2.6, the 'id' property "shall be unique (across the scope of the host server) instance identifier for a specific instance of the Resource". Since these servers are hosted on the same device, their 'id' can't be the same.

ttzeng:
@ichbinblau when I reviewed the code I was wondering whether the code uses the 'id' property to distinguish the resource types. We should use the 'rt' property for that purpose, the 'id' property is not mandatory, and it should be unique on the hosting server. FYI

ichbinblau:
Hi, @ttzeng. My understanding is that we will use multiple host servers of the same type in the demo. So I suppose the uuid would be different as they are on different sensors. Isn't it true?

ttzeng:
@ichbinblau yes, they are multiple servers running on the same device, so OCF spec defined the 'id' property to identify each resources. @gvancuts @poussa please comment

ichbinblau:
Hi. @ttzeng. In our case, uuid cannot be empty cos we need the uuid to subscribe the data updates from the ocf server. eg. http://<gateway-ip>:8000/api/oic/a/led?di=<di>. If it is optional, there is no way to get the data.

ttzeng:
@ichbinblau how did you get the uuid? take an entrt from the REST API server for example:
{"di":"37323863-3365-6237-2d34-6364642d3463","links":[{"href":"/a/fan","rt":"oic.r.fan","if":"oic.if.baseline"}]}
the deviceId is 37323863-3365-6237-2d34-6364642d3463, and the resourcePath is "/a/fan", the resourcePath must be different among the same "oic.r.fan" resource type.

gvancuts:
Hi, @gvancuts<https://github.com/gvancuts>. I did not use cherry-pick. Here are the steps I followed . I guess it was caused by the second rebase pull which overturned sequence.

$ git remote add upstreamhttps://github.com/01org/SmartHome-Demo
$ git fetch upstream
$ git rebase upstream/master
$ git pull --rebase origin environmental_sensor # updated to the commit 6795a1c<https://github.com/01org/SmartHome-Demo/commit/6795a1c2b22e9db50156f4cdfed217b2899a4d74> cos the commit was not up-to-date.
$ git push origin environmental_sensor

Thanks for showing me the correct steps. I think it should be right now.

Thanks @ichbinblau !

I believe too that it was the ‘git pull –rebase’ that created this. I never considered using ‘git pull –rebase’ to be honest but you made me look into it a bit ☺

I believe you could use it but the sequence would be something like that (note that it replaces ‘git fetch’ and ‘git rebase’, and I also believe the parameters you used were incorrect):

$ git remote add upstreamhttps://github.com/01org/SmartHome-Demo
$ git pull --rebase upstream/master
$ git push origin environmental_sensor

I have not tested this but I think I’ll give it a try next time.

Thanks anyway for fixing this PR!

ichbinblau:
@ttzeng Do you mean that two LED sensors on the same device would have the same uuid and rt (oic.r.led) but different resource paths?

ichbinblau:
@gvancuts I was told again and again not to simply use `git pull` to pull updates cos it would take a lot of efforts to merge code. Therefore, it became a habit to pull with `--rebase` now which caused the mess.  :)

gvancuts:
> @gvancuts I was told again and again not to simply use git pull to pull updates cos it would take a lot of efforts to merge code. Therefore, it became a habit to pull with --rebase now which caused the mess. :)

I think this is a good habit in general but I think what caused the problem here is that the wrong arguments were used and it also duplicated what I had asked you to do with 'git fetch' and 'git rebase'. I'm pretty sure (but untested) that 'git pull --rebase upstream/master' would have done the trick... I'll try some other time as it would be a simpler sequence to remember :)

ttzeng:
@ichbinblau I don't know how the iot-rest-api-server generates the uuids, but from OCF spec, servers provide properties such as 'rt' for resource type, and 'id' for unique identifier. Since a device can host multiple resources, each resource may have multiple resource types, so the 'id' property must be able to identify any of them. So, yes, the 'rt' will be the same for all resources of the same type, no mater they are on the same device. If they are on the same device (e.g. smart power plug with multiple binary switches), it seems 'id' would be the only way to identify.

gvancuts:
Disclaimer: I haven't read the complete specs and I'm a little lost as to what the problem is.

Having said that, I think the situation is as follows:
- a 'rt' identifies the type of resource, so multiple OCF servers representing one (or multiple) LEDs would have that 'rt' property set to the same value (e.g. 'led')
- the UUID and 'id' are different. IIRC the UUID is generated by iotivity-node and made persistent using a local dat file, see [Device ID] for a little more details.

I believe what the cloud portal should do is use the 'rt' property to understand what kind of device it is dealing with and use the UUID to uniquely identify each of those.

Does that help/make sense?

If you want to read more about device 'id' and UUID, the following thread (triggered by one of the iotivity-node developer) is enlightening: [Device id?](https://lists.iotivity.org/pipermail/iotivity-dev/2015-December/003201.html)

[Device ID]: https://github.com/otcshare/iotivity-node#provisioning-and-device-id-persistence

ttzeng:
I believe not only the UUID (or the 'di' from iot-rest-api-server) but also the resource path (or 'href' from the api server) are required to unique identify a resource, like [this fragment code](https://github.com/01org/SmartHome-Demo/blob/iotivity_1.2.0/gateway/gateway-server.js#L235-L239) in the gateway. FYI.

ichbinblau:
@gvancuts @ttzeng 
- I can understand that 'rt' ought to be used as sensor type identifier
- But telling from Tonny's Json data, the 'di' I got from the rest api cannot be used to uniquely identify the sensor resource.  
    If that's the case, does it mean that we cannot simply use the 'di' property to differentiate sensors of the same type in the web UI? If so, would it be too clumsy to put both the 'di' in UUID format and resource path in the UI to identify the resource? 

-----
Here is the Json data for LED from Tonny:  
    {"di":"63393562-3338-3764-2d30-6134342d3438","links":[{"href":"/a/led0","rt":"oic.r.led","if":"oic.if.baseline"}]
    {"di":"63393562-3338-3764-2d30-6134342d3438","links":[{"href":"/a/led1","rt":"oic.r.led","if":"oic.if.baseline"}]

ttzeng:
@ichbinblau to be clear, we should use 'di' + 'href' to identify the resources, instead of 'di' + 'rt'. And, in client implementations, we may use any mechanism to filter out the repeated resources (ref [the gateway code to detect duplicate resources](https://github.com/01org/SmartHome-Demo/blob/iotivity_1.2.0/gateway/gateway-server.js#L235-L238)).

Regarding the duplicate entries in my Json data, it may because [a known bug](https://github.com/01org/iot-rest-api-server/issues/41) in the iot-rest-api-server prebuilt in Ostro images used in my setup, the above duplicate detection mechanism should be able to avoid.

Per the [persistence device Id discussion](https://lists.iotivity.org/pipermail/iotivity-dev/2015-December/003201.html) which @gvancuts mentioned, the device Id can be fixed in the secured mode, the device Id alone is not sufficient to distinguish resources on the same device.

ichbinblau:
@ttzeng Thanks for the correction. It's a typo.

nagineni:
I remember we discussed several times earlier that the unique identifier for an OCF resource is the UUID of the device (i.e di) that hosts the resource, and the URI path of the resource (i.e href). 

It was specified in the OCF JS API spec also and I implemented the gateway client based on it.

ichbinblau:
Hi, @gvancuts @ttzeng.  I've almost completed refactoring the code for the misunderstanding.
As the device id is not the unique key to identify a resource, what shall we use to distinguish them? Shall we use the combination of device_id+resource_path or simply the resource_path? Since the resource_path is configurable, it is possible to make it unique manually.

ichbinblau:
Hi, @ttzeng  I have updated the code as we discussed. (use di+resouce path) Would you have a test? Thank you!

ttzeng:
I've tested the the code in the zip file of [this commit](https://github.com/01org/SmartHome-Demo/tree/e303790a17f341c30fadc6abd790cdca6c4d052e), here are the observations:
(1) the docker image can be built and run OK. Change the address of the rest-api-server in the Admin portal OK.
(2) after restarting the docker image, though the web portal can show multiple resources of the same time (i.e. 'LED' of 'oic.r.led'), but they are not in a group (see this screenshot
screenshot from 2016-12-15 12-14-23
) -- Is this the expected behavior?
(3) I have total 5 LED servers running in the same network, but they may not be shown completely most of the time. It looks like the rest-api-server can report the discovered servers completely (see this resource list from the server
[rest-server-report.txt](https://github.com/01org/SmartHome-Demo/files/653780/rest-server-report.txt)
), but the portal may not show all of them. The docker log
[docker-disappear.txt](https://github.com/01org/SmartHome-Demo/files/653786/docker-disappear.txt)
 is attached for your inspection.
(4) The servers will be disappear from the web portal, this screenshot 
screenshot from 2016-12-15 12-17-40
captured when that happened. You could notice from the time embedded in the screenshot filenames, it happened few minutes after last screen captured. Sometimes it may recover, sometimes it completely disappear until restart the docker. The above log was captured when the web portal lost the servers forever, and this log
[docker-recoverable.txt](https://github.com/01org/SmartHome-Demo/files/653794/docker-recoverable.txt)
 captured when the lost servers can be shown again.
(5) The shown resource names can be "append" by user editing, but they can't be changed. For example, it can change 'LED' to 'LED1', but not 'LED' to 'My office LED'. I think this should be a bug.

ichbinblau:
@ttzeng Thanks for the report. 
2)  The LEDs would not be in a group but when you mouse over the card, it should show the uuid+path of the resource.
3) It depends on whether all the resources can report the stream data. I will send you a script by email to test whether each of the resource sends stream data.
4) Would you make sure that the gateway always return a complete list of resources? I can also send you a script to observe this. Cos the portal would only pull the resource list from the gateway server, the sensors on portal disappear when they are gone from the list. 
5) The design is based on our talk in previous meeting. The sensor type would be a fixed prefix for each sensors. @gvancuts

gvancuts:
> 5) The design is based on our talk in previous meeting. The sensor type would be a fixed prefix for each sensors. @gvancuts

This is correct, we can treat that as a feature request but this is the expected behaviour at this point. My opinion is that this is sufficient for a demo but we can certainly open the discussion

ttzeng:
@ichbinblau if the container rely on the 'id' property in the server report for a specific resource, then you may hit [an upstream bug](https://github.com/01org/iot-rest-api-server/issues/43).

ichbinblau:
Hi, @ttzeng. Thank you so much for the test and the finding. 
Yes, the portal code uses the href+di to identify the resource and pulls the data from the resource for the time being. 
I have been using http://\<gateway-ip\>:8000/api/\<href\>?di=xxxxxxx&obs=1 to observe the data continuously. 
If the api does not respond correctly, the sensor would not show up in the UI.  

@gvancuts If it is an upstream bug of the iot rest api, shall we make sure that the tagging feature and the basic feature work and verify the multiple sensor support function when the upstream bug is fixed?

ttzeng:
@ichbinblau how do you handle the response of http://<gateway-ip>:8000/api/<href>?di=xxxxxxx&obs=1? the 'id' property of the response is wrong if there are multiple resources on the same type presented, will that impact the current portal implementation?

ichbinblau:
@ttzeng the code would not validate the 'id' property of the response. As long as http://\<gateway-ip\>:8000/api/\<href\>?di=xxxxxxx&obs=1 gives back response, it reads the value straight away.

ttzeng:
@ichbinblau then you should have no impact by [this bug](https://github.com/01org/iot-rest-api-server/issues/43). The given container image still can't show the complete list of resources, though the missing resources do have representations in the `/api/oic/res` report, and the query of `href?di=xxx` does return the property of resource 'xxx'

BTW, since the portal uses `http://<gateway-ip>:8000/api/<href>?di=xxxxxxx**&obs=1** ` to observe the resource updates, what if a specific resource was not included in the report of `/api/oic/res` but appear later?

gvancuts:
I have lost a bit the flow of the conversation. Is there any open question I should respond to myself?

Do drop me a note when this could be merged into the main repo! ;)

ichbinblau:
Hi, @gvancuts Tonny was able to connect me to his physical gateway this Monday.  After some troubleshooting on Tuesday, I found that I was unable to fetch data from 5 out of 14 resources in obs mode which may result in the incomplete sensors list in Web UI.  And the total number of sensors in the gateway may fluctuate at times which made the sensors disappear.  I think Tonny would verify the image again when he is back after fixing the environment issue.

ttzeng:
I am wondering it's the environment issue, as the companion app (using native IoTivity) can access all the servers, it may be an issue with the iot-rest-api-server that @ichbinblau got only part of servers through the API sometimes. Or, is it possible to implement some error recovery mechanism to make the web portal robust?

ichbinblau:
The Cloud portal only interacts with rest api. In that case it may not behave the same way with the companion app. 
Sure, I can improve it in future. But the purpose of this PR is to support the new env sensor, tagging feature and multiple sensors of the same type.  

The env sensor cannot be reached thru the rest api. ( I guess a rest api restart may solve the problem. )
untitled picture

The tagging feature seems already work here. And the multiple sensors have also been supported. 
untitled picture1

Shall we make sure the new features function first?

gvancuts:
I think we need to understand if anything odd is happening within the iot-rest-api-server before we add some extra logic in the cloud to make it more robust. The key would be to understand if it's an expected behaviour that we are seeing or something that is failing. Is there any extra datapoints you could collect from the iot-rest-api-server (by running it in verbose mode for example). @poussa @nagineni any further suggestion on how we should go about debugging this?

Once we have reasonable confidence that the new features are functional, we could merge this patch and file specific issues if some things are unstable or broken.

gvancuts:
@ttzeng , what should I do with this PR?

gvancuts:
Hi @ichbinblau , I have spoken with @ttzeng  and he was thinking he would be good to re-test this after PR #90 has been merged, we suspect many of the problems we saw are solved there and it would give us a better baseline to test this. Is that doable from your perspective? (also keeping in mind that some extra modifications need to be made to the cloud to support this update to IoTivity 1.2.0-1)

ichbinblau:
Hi, @gvancuts. I am good with a re-test. Yeah, I saw the PR #90 from Sakari and Babu. I am working on the cloud modification now.

gvancuts:
Thanks @ichbinblau

PR #60, opened 2016-11-23
documentation: improve environmental-sensors README.md

wanghongjuan:
- Fix typos issue
- Add detail steps for setting up dev environment
- Add description that how to enable BLE on gateway

Signed-off-by: Hongjuan Wang <hongjuan.wang@intel.com>

### DIFF
--- a/sensors/environmental-sensors/README.md
+++ b/sensors/environmental-sensors/README.md
@@ -18,7 +18,7 @@ describes how to wire these sensors to the Arduino 101 board.
 
 Both sensors operate at 3.3V or 5V. The BME280 uses I2C to communicate with
 the Arduino, while the UV sensor sends data via analog port (ADC). For the
-BME280, its SDL, SDA and GND pins are connected to the SDL, SDA and GND pins
+BME280, its SCL, SDA and GND pins are connected to the SCL, SDA and GND pins
 respectively, and the Vin pin is connected to 5V pin on the port. The Grove
 UV sensor has 3 pins, GND to GND pin, Vcc to 3.3V pin and SIG pin to A0 pin
 on the board.
@@ -35,20 +35,30 @@ configuration options from the arc/proj.conf file.
 The Grove LCD communicates with the sensor subsystem through the I2C bus. When
 connecting the Grove LCD to the Arduino 101, either directly or through a Grove
 Base Shield, you need to make sure that the I2C SDA and SCL lines have pull-up
-resistors connected between GND and the 3.3V power source.
+resistors to 3.3V.
 
 Take note that even though SDA and SCL are connected to a 3.3V power source, the
-Grove LCD VDD line needs to be connected to the 5V power line, otherwise
+Grove LCD VCC line needs to be connected to the 5V power line, otherwise
 characters will not be displayed on the LCD (3.3V is enough to power just the
-backlight).
+backlight). If you want to connect the LCD through a Grove Base Shield, please make sure
+that the base shield has a slide switch for selecting 5V or 3.3V as VCC, then select
+5V as VCC.
 
 The Grove LCD can be found here:
 https://www.seeedstudio.com/Grove-LCD-RGB-Backlight-p-1643.html
 
 ## Building and flashing the app to Arduino 101
 In order to build the Zephyr app for the Arduino 101 board, first setup the
-Zephyr dev environment following the guidance here.
-* https://wiki.zephyrproject.org/view/Arduino_101
+Zephyr dev environment following steps:
+* Setup dev environment on Host following this [guide](https://www.zephyrproject.org/doc/getting_started/getting_started.html)
+* Download Zephyr project source code which is [supported](#supported-zephyr-versions) and enable environment configure
+```
+    $ git clone https://gerrit.zephyrproject.org/r/zephyr && cd zephyr && git checkout tags/[supported_version]
+    $ source ./zephyr-env.sh
+```
+* Flash a bluetooth firmware named [`MyNewt`](https://wiki.zephyrproject.org/index.php?title=Arduino_101&oldid=975#Bluetooth_firmware_for_the_Arduino_101) which is supported by Zephyr
+
+**Note**: You can learn more information from this [wiki](https://wiki.zephyrproject.org/index.php?title=Arduino_101&oldid=975)
 
 In the app source folder, use the following commands to build and flash the app
 to Arduino 101 board:
@@ -70,15 +80,20 @@ to Arduino 101 board:
 ## Data flow and usage
 The sensor app uses standard BLE's ESS (Environmental Sensing Service) service and attributes to send data. It acts as a BLE peripheral device advertising itself as a "Zephyr Environmental Sensor" that any client can scan, connect and have the data via "read" or "notify" methods. On Android or iOS devices with BLE support, BLE apps like nRF Connect can be used to communicate and receive sensor data. In the SmartHome-Demo system, a remote OCF resource server using BLE connects, receives data from the sensor and shares to other devices in the same network.
 
-The OCF resource server is implemented in "ocf-servers/js-servers/environmental_sensor.js". Use the following command to launch it:
+The OCF resource server is implemented in `ocf-servers/js-servers/environmental_sensor.js`. Use the following command to launch it:
 ```
     $ node <path>/environmental_sensor.js
 ```
 
 Please note that it requires noble to run. noble can be installed by using following command in the SmartHome-Demo folder:
 ```
-    $ npm install noble
+    $ npm install noble@^1.6.0
 ```
+Please make sure that BLE is enabled on the device running the OCF resource server(i.e. `ocf-servers/js-servers/environmental_sensor.js`), you can enable it via below command:
+```
+    # rfkill unblock bluetooth
+```
+Use command `hciconfig` to check if it is unblocked and running.
 
 ## Supported Zephyr versions:
 The code has been verified to work with Zephyr v1.4.0 and v1.5.0


wanghongjuan:
@kiendinh Since the guide that you linked about setting up Zephyr dev environment is updated frequently for Zephyr project, not suite this sample, so I update the description. Could you please help to review this PR?

sensors/environmental-sensors/README.md
```diff
@@ -47,8 +47,16 @@ https://www.seeedstudio.com/Grove-LCD-RGB-Backlight-p-1643.html
...
-Zephyr dev environment following the guidance here.
-* https://wiki.zephyrproject.org/view/Arduino_101
+Zephyr dev environment following steps:
+* Setup dev environment on Host following this [guide](https://www.zephyrproject.org/doc/getting_started/getting_started.html)
+* Download Zephyr project source code and enable environment configure
+```
+    $ git clone https://gerrit.zephyrproject.org/r/zephyr && cd zephyr && git checkout tags/[supported_version]
+    $ source ./zephyr-env.sh
+```
+* Flash a bluetooth firmware named [`MyNewt`](https://wiki.zephyrproject.org/index.php?title=Arduino_101&diff=976&oldid=975#Bluetooth_firmware_for_the_Arduino_101) which is supported by Zephyr
```
gvancuts: That URL points to the (right) section of a diff between two versions of the wiki page. We should either point at the same section in the latest rev of the doc or point at a specific rev (e.g. https://wiki.zephyrproject.org/index.php?title=Arduino_101&oldid=975#Bluetooth_firmware_for_the_Arduino_101)

sensors/environmental-sensors/README.md
```diff
@@ -47,8 +47,16 @@ https://www.seeedstudio.com/Grove-LCD-RGB-Backlight-p-1643.html
...
+
+**Note**: You can learn more information from this [wiki](https://wiki.zephyrproject.org/index.php?title=Arduino_101&diff=976&oldid=975)
```
gvancuts: Same comment, this points at a diff. Can we replace this with the latest rev or a specific one?

sensors/environmental-sensors/README.md
```diff
@@ -79,6 +87,11 @@ Please note that it requires noble to run. noble can be installed by using follo
 ```
     $ npm install noble
 ```
+Meanwhile, please make sure BLE is enabled on the OCF resource server side, if you are using Liunx* based OS as gateway, you can enable it via below command:
+```
+    $ rfkill unblock bluetooth
+```
+Use command `hciconfig` to check if it is unblocked and running.
```
gvancuts: Is this available by default on Ostro?

sensors/environmental-sensors/README.md
```diff
@@ -79,6 +87,11 @@ Please note that it requires noble to run. noble can be installed by using follo
 ```
     $ npm install noble
 ```
+Meanwhile, please make sure BLE is enabled on the OCF resource server side, if you are using Liunx* based OS as gateway, you can enable it via below command:
```
gvancuts: The board that will run the OCF resource server is not necessarily the gateway so I would suggest a slightly different wording here: "Please make sure that BLE is enabled on the device running the OCF resource server (i.e. `ocf-servers/js-servers/environmental_sensor.js`). If you are using a Linux-based OS, you can enable it via below command:"

sensors/environmental-sensors/README.md
```diff
@@ -79,6 +87,11 @@ Please note that it requires noble to run. noble can be installed by using follo
...
+```
+    $ rfkill unblock bluetooth
```
gvancuts: I guess you need 'sudo' here? (or perhaps use the '#' to indicate the login prompt is from the root user)

sensors/environmental-sensors/README.md
```diff
@@ -79,6 +87,11 @@ Please note that it requires noble to run. noble can be installed by using follo
 ```
     $ npm install noble
 ```
```
gvancuts: While you're making modifications to this file (thanks for doing it by the way!). Can you also add the following enhancements:
1. Add a version of noble that is known to work (the one we've tested with)
2. A few lines above this one in the same file, replace "ocf-servers/js-servers/environmental_sensor.js" by `ocf-servers/js-servers/environmental_sensor.js` (so it gets formatted as code text)

sensors/environmental-sensors/README.md
```diff
@@ -47,8 +47,16 @@ https://www.seeedstudio.com/Grove-LCD-RGB-Backlight-p-1643.html
 
 ## Building and flashing the app to Arduino 101
 In order to build the Zephyr app for the Arduino 101 board, first setup the
-Zephyr dev environment following the guidance here.
-* https://wiki.zephyrproject.org/view/Arduino_101
+Zephyr dev environment following steps:
+* Setup dev environment on Host following this [guide](https://www.zephyrproject.org/doc/getting_started/getting_started.html)
+* Download Zephyr project source code and enable environment configure
```
gvancuts: Can we add a note pointing at the bottom of the page where the supported versions are mentionned?

sensors/environmental-sensors/README.md
wanghongjuan: Yes, I will update the description.

sensors/environmental-sensors/README.md
wanghongjuan: Yes, I have tested it on Joule 470x with Ostro-xt#194.

gvancuts: Thanks for confirming this point.

wanghongjuan:
Hi @gvancuts I have updated the file, PTAL. Thanks a lot.

sensors/environmental-sensors/README.md
```diff
@@ -38,7 +38,7 @@ Base Shield, you need to make sure that the I2C SDA and SCL lines have pull-up
 resistors connected between GND and the 3.3V power source.
```
ttzeng: I think it's "pull up resisters to 3.3V"? "pull up resisters connected between GND and the 3.3V power source" sounds a voltage divider?

ttzeng: If through a Grove base shield, the base shield has a slide switch for selecting 5V or 3.3V as Vcc, thus the wiring will be different. IMHO.

sensors/environmental-sensors/README.md
```diff
@@ -47,8 +47,16 @@ https://www.seeedstudio.com/Grove-LCD-RGB-Backlight-p-1643.html
 
 ## Building and flashing the app to Arduino 101
 In order to build the Zephyr app for the Arduino 101 board, first setup the
-Zephyr dev environment following the guidance here.
-* https://wiki.zephyrproject.org/view/Arduino_101
+Zephyr dev environment following steps:
+* Setup dev environment on Host following this [guide](https://www.zephyrproject.org/doc/getting_started/getting_started.html)
+* Download Zephyr project source code which is [suppored](#supported-zephyr-versions) and enable environment configure
```
ttzeng: Please make sure the link of the "supported" is valid?

sensors/environmental-sensors/README.md
```diff
@@ -70,15 +78,20 @@ to Arduino 101 board:
...
 ```
     $ node <path>/environmental_sensor.js
 ```
 
 Please note that it requires noble to run. noble can be installed by using following command in the SmartHome-Demo folder:
 ```
-    $ npm install noble
+    $ npm install noble@1.7.0
+```
+Please make sure that BLE is enabled on the device running the OCF resource server(i.e. `ocf-servers/js-servers/environmental_sensor.js`). If you are using a Liunx-based OS, you can enable it via below command:
```
ttzeng: "a Linux based OS" sounds redundant as the all parts of the demo more or less imply using Linux.
And, per this [post](http://rexstjohn.com/configure-intel-edison-for-bluetooth-le-smart-development/), it seems below command should be enter every time, or permanently disable the bluetoothd with systemd?

kiendinh:
Thank you for doing this, @wanghongjuan. It's really appreciated.

sensors/environmental-sensors/README.md
wanghongjuan: We can get same result as add pull-up resistor when using Grove base shield and switch to 5V as VCC, if using 3.3V as VCC, the RGB LCD also can show data, but no backlight.

sensors/environmental-sensors/README.md
wanghongjuan: I try to enable BLE on Edison following the post that you mentioned, it doesn't work, but the BLE can work well on Joule.  I heard that there are problems with BLE on Edison from @gvancuts .

sensors/environmental-sensors/README.md
ttzeng: Yes, what? will you rephrase this sentence?

ttzeng: Did you try cold boot the LCD module and still 100% function well? I meant we may put a note for the slide switch if the Grove base shield is used.

wanghongjuan: >> I think it's "pull up resisters to 3.3V"? "pull up resisters connected between GND and the 3.3V power source" sounds a voltage divider?

Pull-up is the uncertainty of the signal through a resistor clamp in the high-level, while the role of resistance from the current limiting. And I will update the sentence from "pull-up resisters connected between GND and the 3.3V power source" to "pull-up resisters to 3.3V"

wanghongjuan:
Hi, @gvancuts @ttzeng @kiendinh I have updated the doc, PTAL. :)

gvancuts:
👍

gvancuts:
@ttzeng @kiendinh , any further comments?

kiendinh:
@gvancuts @wanghongjuan , the only thing left is the version of noble. Since it's already been specified in the package.json, and will be installed for the whole project by running "npm install" at the top level folder, specifying it here will be difficult to be in coherence with the remaining parts. Thanks.

wanghongjuan:
@kiendinh Following what you said, my understanding is this demo must be depent on the `gateway` to run, since the `package.json` that you mentioned is exist in folder `gateway`. Is it right?
BTW, I noticed that this [tutorial](https://01.org/smarthome) is based on Ostro image to setup this demo. The guide only instruct users to install `mraa` and `socket`, for `iotivity-node`, it has been built into this image, there is no need to install `iotivity-node` by running ` npm install`, and it seems that `http` and `express` is not necessary,. If I want to use a Android control app([`smart-home-control`](https://github.com/siovene/smart-home-control)) which using Cordova OCF plugin to verify the running OC resources in the connected Smart Home instead of the gateway, the `package.json` will be useless. How do you think about this? Thanks. :)

kiendinh:
@wanghongjuan For most of the devices user need to checkout the SmartHome-Demo code, and it's reasonable to keep the consistency in the code - not hardcoding here and there. For the Android app, as you worried, it doesn't even require the sensor to be connected over BLE to test the OCF. There is a default simulation mode if BLE not found.
Updating the tutorial is another story. We'll need to get it done sometime.

ttzeng:
@wanghongjuan , regarding the tutorial, we didn't instruct users to install the dependency modules by "npm install", so the package.json is not used and contains out-of-date dependencies such as 'http'.

wanghongjuan:
@kiendinh First, as README said, `In the SmartHome-Demo system, a remote OCF resource server using BLE connects, receives data from the sensor and shares to other devices in the same network.`  If the guide doesn't have description about `noble` and I don't know how to enable it, then users need to get answer via check the OCF resource server code, internet or report issues etc. But I think keep the description will be better. Second, the data which transfer from the simulation mode isn't true, and the data shows in LCD is mismatch with Android app. Following the data on Android app, it can not make sure whether the sensor is working. It doesn't make sense.

wanghongjuan:
Ohh, May be my understanding is wrong,  problem is version of noble that you mentioned, I can update it align with the `package.json` now

kiendinh:
Yep, it's just about the version of BLE that should be used. And it's already said v1.6.0 or later in the package.json.

By the way, a recent change moved the package.json to the gateway folder that may cause the confusion.

@gvancuts, can we move package.json back to the top level folder? Since other ocf-resource servers like fan.js etc are also relying on the iotivity-node, it would be easier to have the package.json in the top level folder, especially when we try to use the demo on platforms other than Ostro, which don't have iotivity-node by default.

ttzeng:
Babu have moved the package.json back to the root of the repo in the 1.2.0 branch, FYI

gvancuts:
Wouldn't it be better to have two package.json files, one in each folder that may need it? I'm assuming there will be common dependencies but also areas where they diverge?

I personnally find having a package.json in the root folder a little confusing.

ttzeng:
IMHO, if we view the entire repo as a project that can be switch to either OCF servers or clients, then I think it makes sense to 'npm install' the dependencies after "repo clone"; if we view the repo contains two modules, then "npm install" under each module also making sense.

Either way is fine for me, but I think no harm to install "redundant" node modules (e.g. 'websocket', etc.) on a platform which will only run OCF servers. In some situation, we may want to run the gateway-server for backup. (the iot-rest-api-server module has been built-in in bother gateway and edison images are a typical example)

gvancuts:
OK, let's move the file to the root folder (and make any adjustment needed at the same time). Can one of you prepare a PR for this?

wanghongjuan:
Hi, @gvancuts I have updated the version of `noble` align with the `package.json` and submitted a new PR #67 for moving the file to root folder. PTAL. Thanks. :)

gvancuts:
Thanks @wanghongjuan 

@kiendinh @ttzeng , any further comments from you?

gvancuts:
Merging, we can address additional feedback with new PRs if needed

PR #61, opened 2016-11-23
Dockerfile: update file to enable automatic builds for PR testing

gvancuts:
In its current form, the Dockerfile makes it impossible to have
DockerHHub automatically build images to test GitHub Pull Requests
for the cloud portal code. This is because the code is pulled inside
the container using "git clone https://github.com/01org/SmartHome-Demo"
but the changes are not yet in that repo.

The solution proposed is to move the Dockerfile from its current
location (in smarthome-web-portal/tools/docker to smarthome-web-portal)
so that the entire content of the folder is part of the context when
building the container and the code, which includes the latest
modifications, is now copied to the container.


gvancuts:
If you want to test such image, pull it using: docker pull gvancuts/smarthome-demo-pr-testing:automated-builds-for-dockerhub

ichbinblau:
Hi, @gvancuts . I think this is the right way.  The project folder should be under the same directory with the Dockerfile. In that case, we could use COPY instead of git clone to pull the latest source code. I have verified the image pulled from your Dockerhub repo and both portals work.  Thank you for providing the solution.

ttzeng:
I've validated this PR by applying the proposed patch on current master HEAD (91a10407b6a95c46af6e218bf9e8f00b146ebd41). The docker image can be built and function as usual. Please proceed with the merge.

gvancuts:
Thank you @ichbinblau @ttzeng

PR #64, opened 2016-12-05
sensors: LCD display: enabled displaying UV index and fixed a typo

kiendinh:


gvancuts:
This looks good to me 👍 

There is another typo [here](https://github.com/01org/SmartHome-Demo/blob/master/sensors/environmental-sensors/arc/src/main.c#L161) you may want to include in this comit: `s/humidity/humidity`

@ttzeng , are you OK to merge without testing (this looks simple enough to me but I'd like your opinion)

kiendinh:
Thank @gvancuts! I just re-pushed with a fix for that typo.

gvancuts:
Thanks @kiendinh 👍 
@ttzeng , what do you say?

sensors/environmental-sensors/arc/src/main.c
```diff
@@ -158,7 +158,7 @@ void main(void)
 			temp.val2/100000, 223 /* degree symbol */);
 		glcd_print(glcd, row, strlen(row));
 
-		/* display himidity on LCD */
+		/* display humidity on LCD */
 		glcd_cursor_pos_set(glcd, 17 - strlen(row), 0);
```
gvancuts: @ttzeng all good on your side now?

qiuzhong:
The Pressure + UV could be more than 16 characters. I build the apps based on this PR. But I got this on the Grove LCD v2:

0 | 1 | 2 |  3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
T | : | 2 | 3 | . | 0 | ° | C |  |  |  | R | H | : | 5 | 0
P | : | 1 | 0 | 4 | . | 2 | 9 | 5 | k | P | U | V | : |  2 | 5 

I tried replacing 17 to 16, rebuilt the apps and got this:

0 | 1 | 2 |  3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
T | : | 2 | 3 | . | 0 | ° | C |  |  | R | H | : | 5 | 0 | %
P | : | 1 | 0 | 4 | . | 2 | 9 | 5 | k | U | V | : | 2 | 5 | 5 

I think the complete data to be displayed should be:
T: 23.0 °C RH: 50%
P: 104.295kPa UV:255

0 | 1 | 2 |  3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
T | : | 2 | 3 | . | 0 | ° | C |  | | | | R | H | : | 5 | 0 | %
P | : | 1 | 0 | 4 | . | 2 | 9 | 5 | k | P | a | U | V | : | 2 | 5 | 5 

Obviously, the UV index value is wrong, but the bluetooth client can get the correct UV index: 1. I don't know how this happened.
But even the UV index value is valid (generally less than 11), one byte more is required.

kiendinh:
Thank you, @qiuzhong. I just re-pushed a new to fix the UV value display. Please give it a try.

qiuzhong:
The UV index value display in the Grove LCD is still not good. This time I got:

0 | 1 | 2 |  3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
T | : | 2 | 2 | . | 3 | ° | C |  |  |  | R | H | : | 4 | 1
P | : | 1 | 0 | 5 | . | 3 | 3 | 6 | k | P | U | V | : |  2 | 5

BTW, I think the start positions the humidity and UV strings need to adjust, cause the last character (for example percentage symbol for the humidity value) is chopped.

kiendinh:
@qiuzhong, thanks. I've re-pushed a new one. Please take a look.

qiuzhong:
I just rebuilt the app. And now it can display the humidity value with the percentage symbol completely. 

0 | 1 | 2 |  3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
T | : | 2 | 2 | . | 3 | ° | C |  |  | R | H | : | 4 | 1 | %
P | : | 1 | 0 | 5 | . | 2 | 5 | 3 | k | U| V | : | 2 | 5 | 5 

Though the "Pa" in unit is gone.
You assume the fractional part of pressure value number to be 2 bytes in unit kPa. But it could take 3 bytes.

qiuzhong:
Hi, @kiendinh, I print the uv_signal in the app and capture it from the serial port.
Its value is about 200~215, is this in the valid range of the UV sensor?
After applying this value to your algorithm to compute the UV index, I find uv_index is already
a negative number: -1, which happens to be 255 in the uint8_t form. Perhaps it can explain why the Grove LCD displays it as 255.

kiendinh:
@qiuzhong Please check the latest patch then. Thanks.

qiuzhong:
I applied to the latest patch. This time the UV index seems valid (0, in doors ). And the Grove LCD displays well.

0 | 1 | 2 |  3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
T | : | 2 | 2 | . | 2 | ° | C |  |  | R | H | : | 4 | 8 | %
P | : | 1 | 0 | 4 | . | 9 | 8 | 9 | k | P| a | U| V | : | 0

However, there is another problem: both the bluetooth client app and power-uart.js OCF server receive the UV index value as 1 from the Arduino 101. I remember that the UV index value was recevied as 1 too before this patch.

kiendinh:
@qiuzhong, it's been fixed now. Thanks.

qiuzhong:
@kiendinh , both the bluetooth client app and power-uart.js OCF server received the UV index value as 0 from the Arduino 101 in my test. It's in consistent with the LCD. The environmental sensor can work well now! Thank you!

kiendinh:
@qiuzhong, you can bring the device to a window or outside under the sun to test. It works pretty nicely measuring the UV. There are several ways to calculate the index. We're using the one that's being used by the mraa library. It has some limits though but I hope that we can demonstrate the concept.

@gvancuts, could you please merge the PR? Thanks.

gvancuts:
Thanks everyone, merging.

PR #87, opened 2017-01-23
power-uart.js: Fixed the format of power consumption values.

kiendinh:
They should be in mW instead of Watt when receiveing via BLE.


qiuzhong:
@kiendinh, I don't get it. 
Before applying to this PR, I can get the consistent powe1 and power2 values from the LCD and power-uart.js terminal output.
Now the terminal output values are 1000 times the amount of the LCD ones.

qiuzhong:
In the DC power meter app, the power1 and power2 values (in uW unit) are sent to x86 core without any unit conversion. And later they're converted into mW by dividing them by 1000 for the LCD display. 

I didn't see any unit conversion either before the x86 core sends them out via the Bluetooth service. 
I don't understand why this conversion is dropped on the OCF server side.

kiendinh:
Thank you @qiuzhong . That's a terrible mistake. There was a bug that value from the AC power meter, ie 819204 mW, was displayed as 0.819 W in the cloud. Please let me close this PR for further consideration.

PR #89, opened 2017-01-23
power-uart.js and DC power meter: Fixed the value format

kiendinh:
Power consumption value was not transferred correctly and thus
caused misinterpretation in the cloud portal. Fixed now.


gvancuts:
@qiuzhong , could you please verify this (unless I got this very wrong, this is the follow-up to #87).

qiuzhong:
Verified.  It works well for the Zephyr 101 board and the OCF server power-uart.js (mW as the unit of power). And the web portal can display correct consumption value as well while it uses Watt as the unit of power.

gvancuts:
Thank you @qiuzhong , merging this.

PR #83, opened 2017-01-17
Virtual SmartHome in a Docker container

gvancuts:
This is the initial implementation of a Docker container that
provides a virtual (simulated) SmartHome that includes the following
components:
* IoT REST API Server
* Home GW
* OCF servers (simulated)

Signed-off-by: Geoffroy Van Cutsem <geoffroy.vancutsem@intel.com>

PR #90, opened 2017-01-27
Update OCF servers and client implementation to latest JS spec and IoTivity 1.2.0-1

nagineni:
This branch includes the following changes:

- Updated OCF servers implementation to latest JS spec and IoTivity 1.2.0-1
- Made rgb_led server to notify the updated properties even no mraa module installed(this helps us to test the update functionality in desktop env).
- Updated gateway-server implementation to latest OCF JS spec and IoTivity 1.2.0-1.
- Added AUTHORS.txt
- Removed presence since it is no longer defined by the OCF specification. Global presence is now enabled by default in the bindings.
- Updated iotivity-node dependency to v1.2.0-1.
- Added changes to exit OCF servers gracefully.
- Removed security related stuff in this version of the demo. We will enable it when IoTivity turns the SECURED flag ON.


nagineni:
@poussa @uhofemei @gvancuts PTAL

nagineni:
I tested all these changes locally and also on custom Yocto build wiith iotivity  v1.2.0. REST API server also updated to iotivity v1.2.0 here: https://github.com/01org/iot-rest-api-server/pull/46

package.json
```diff
@@ -1,17 +1,24 @@
...
     "test": "echo \"Error: no test specified\" && exit 1"
   },
-  "author": "",
   "license": "Apache-2.0",
   "dependencies": {
-	"iotivity-node": "^1.1.1-3",
-	"express": "*",
-	"websocket": "*",
-	"noble": "^1.6.0"
+    "iotivity-node": "^1.2.0-1",
```
gvancuts: Why not directly expressing a dependency on iot-rest-api-server, like I was proposing in #83 ?

nagineni: We do not need REST API server for OCF servers and client(i.e rules engine/3D UI).

Adding dependency only on REST server might break the OCF server and client startup, if someone is using npm@2 to install the modules. npm@2 installs the modules in nested directory (like SmartHome-Demo/node_modules/iot-rest-api-server/node_modules/iotivity-node) and that will cause an error saying cannot find module 'iotivity-node' when we do `require('iotivity-node')` .

This may not be the issue with npm@3 since it installs the nested dependencies in the top-level node_modules directory (SmartHome-Demo/node_modules/iotivity-node).

nagineni:
> This looks good to me! Does that need any additional changes to the following parts:

> Zephyr-based sensors?

I have not tested Zephyr-based sensors, but no changes required AFAIK.

> Cloud portal?

Cloud portal needs to be adapted to the couples chances we made in the REST API server. 

1. Instead of PUT, use POST for update: https://github.com/01org/iot-rest-api-server/commit/77c82498
2. Change http response payload format : https://github.com/01org/iot-rest-api-server/commit/4d9548a
```
Old: {"properties":{"state":false,"power":10}}
new: {"state":false,"power":10}
```

package.json
nagineni: Adding dependency on both iotivity-node and iot-rest-api-server should be okay IMO.

gvancuts:
@kiendinh can you test the Zephyr-based sensors and prepare any adjustments if needed?

@ichbinblau can you prepare a PR to make the necessary adjustments to the portal?

package.json
gvancuts: How would the system react if you specific a version of iotivity-node in this file that's different than the one in the iot-rest-api-server package.json file?

nagineni: There won't be any issues with the module installation, but things may not work as expected if there are any API changes between the versions.
Like now, running client and servers with different version of iotivity-node is not recommended.

gvancuts: In this case, my preference would be to replace the iotivity-node dependency by a dependency to iot-rest-api-server with the right version. I understand this the latter is not needed for the gateway function itself, nor to run the OCF servers but not having it installed breaks the cloud portal and I tend to see this repo as a whole.

If there is a potential problem with some versions of npm, can we express a dependency on a version that works?

uhofemei: I'm hesitant to upgrade the demo given that it's working very stable right now. Does this version bump introduce changes that would affect the overall OCF messaging we've been using so far?

gvancuts: The demo in this repo can live in a loosely coupled fashion with the demo kit. We should strive to not divert too much, nor for too long but things need to move forward. IoTivity 1.2.x was released end of October last year and we'd want to adopt it at some point.

Having said that, we obviously need to test this before merging the code. @nagineni has done a great deal of work testing and adjusting lots of components so the demo does not break. If your testing shows that something is missing, please indicate it clearly. A generic concern, that is more about messaging than technical issues, is not a reason to hold off merging the PR.

gvancuts:
I have discussed this offline with @ttzeng and @kiendinh , we should be good with this PR, I'll merge this now and we can address issues if/when they come up.

@ichbinblau is also preparing a PR for the cloud portal and it will be submitted later (and most likely will need that we re-test, and possibly re-work PR #52 before).

PR #92, opened 2017-01-31
OCF JS servers: add firewall information in the README.md

gvancuts:
Add information on how to disable/re-configure the firewall
when using the Ostro OS as the base OS hosting OCF servers.


gvancuts:
@nagineni @ttzeng can you please take a look

ocf-servers/js-servers/README.md
```diff
@@ -76,6 +76,18 @@ Not having those available will disable the corresponding device and switch to s
...
+Note that when using the [Ostro OS](https://ostroproject.org), its firewall will block the IoTivity network traffic by default. A brute-force approach is to disable the firewall altogether as follows:
+```
+# systemctl disable iptables
+# systemctl disable ip6tables
+```
+
+A better solution would be to open up just the ports that are required. Port 5683 is used by the CoAP protocol during the device discovery phase and other (random) ports are used later to communicate with the device. If you wish to open up just the required ports, you can do so as follows:
+```
+# iptables -A INPUT -p udp --dport 5683 -j ACCEPT
+```
```
nagineni: Port 5684 is used for secure traffic, so worth mentioning that one also here. We will enable secure mode when iotivity turns the SECURED flag on.

Also, add instructions to open ports via ip6tables:

ip6tables -A INPUT -s fe80::/10 -p udp -m udp --dport 5683 -j ACCEPT
ip6tables -A INPUT -s fe80::/10 -p udp -m udp --dport 5684 -j ACCEPT

In case of random ports, the port range can be given like below:
iptables -A INPUT -p udp --dport <start:end> -j ACCEPT
ip6tables -A INPUT -s fe80::/10 -p udp -m udp --dport <start:end> -j ACCEPT

gvancuts:
Thanks @nagineni , I have updated the PR.

nagineni:
@gvancuts thanks!

LGTM

PR #94, opened 2017-02-10
sensors: Added an OCF temperature node as an example

kiendinh:
@poussa, @gvancuts @nagineni @uhofemei @ttzeng: I just added an OCF temperature node example. Please take a look. Thanks


sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+The Grove temperature sensor should be connected to port A0 on the Grove base
+shield.
+
+## Building and flashing the applications
+In order to build the Zephyr app for the Arduino 101 board, first setup the
+Zephyr dev environment following steps:
+* Setup dev environment on Host following this [guide](https://www.zephyrproject.org/doc/getting_started/getting_started.html)
+* Download Zephyr project source code and enable environment configure
+```
+    $ git clone https://gerrit.zephyrproject.org/r/zephyr && cd zephyr
```
gvancuts: Is there any specific version of Zephyr (release or tag) that should be checked out?

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
+# OCF Zephyr Temperature Node
+This is an example of Zephyr temperature sensor functioning as an OCF node on
```
gvancuts: Is 'node' the appropriate terminology here? I would personally prefer that we use 'OCF server' to be more consistent with the rest of the documentation.

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+an Arduino 101 board, using Zephyr and iotivity-constrained. The device collects
+temperature data using a Grove temperature sensor and sends to a connected
+device using 6LowPAN over BLE.
+
+## Sensor wiring
```
gvancuts: Can you add a pointer to the Grove Starter Kit for Arduino or the Grove Shield and temp sensor that you use?

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+    (e.g. echo "connect E6:DF:CD:CE:65:14 2" >/sys/kernel/debug/bluetooth/6lowpan_control)
+
+    # ifconfig  ##You should see bt0 interface, if everything was okay
+```
+
+### Run the tests on Host:
+#### From the browser:
+
+Discovery: http://localhost:8000/api/oic/res  ## You should see the /a/temperature resource.
+Get: http://localhost:8000/api/oic/a/temperature?di<device ID> ## You should see the temperature properties.
```
gvancuts: I think  you are missing an '=' sign in between `di` and `<device ID>`

sensors/ocf_temperature_sensor/x86/src/Makefile
```diff
@@ -0,0 +1,119 @@
+ccflags-y += -I${ZEPHYR_BASE}/net/ip
+ccflags-y += -I${ZEPHYR_BASE}/net/ip/contiki
```
gvancuts: Out of curiosity, does that use the previous IP stack (from Contiki) or the new one? I'm surprised to see such path here.

sensors/ocf_temperature_sensor/x86/src/Makefile
```diff
@@ -0,0 +1,119 @@
...
+ccflags-y += -I${ZEPHYR_BASE}/net/ip/contiki/os/lib
+ccflags-y += -I${ZEPHYR_BASE}/net/ip/contiki/os
+ccflags-y += -I$(ZEPHYR_BASE)/include/drivers
+ccflags-y += -I$(ZEPHYR_BASE)/drivers
+ccflags-y +=-I. -I${IOTIVITY_CONSTRAINED_BASE}/include \
+		-I${IOTIVITY_CONSTRAINED_BASE} \
+		-DDTLSV12 -DWITH_SHA256 -DDTLS_CHECK_CONTENTTYPE -DWITH_OCF \
+		-I${IOTIVITY_CONSTRAINED_BASE}/deps/tinydtls -DNDEBUG
```
gvancuts: Is this really `-DNDEBUG` oir did an extra `N` sneak in?

sensors/ocf_temperature_sensor/x86/src/main.c
```diff
@@ -0,0 +1,144 @@
...
+		val = data;
+		temp = sensor_value_to_double((struct sensor_value *)val);
+		printk("Getting temp sensor data = %d\n", (uint32_t)temp);
+		break;
+	default:
+		break;
+        }
+}
+
+void
```
gvancuts: Extra line break?

sensors/ocf_temperature_sensor/README.md
kiendinh: Thanks. I've updated in the repushed commit.

sensors/ocf_temperature_sensor/README.md
kiendinh: Thanks. I've added the links in the repushed commit.

sensors/ocf_temperature_sensor/README.md
kiendinh: No, it's master branch. It wouldn't compile with the latest 1.6 release. Neither does the Zephyr.js. I would like to use the same revision for both of them. Let's wait a little more to see if the new release meets the timeline and fix the revision with Sakari's team.

sensors/ocf_temperature_sensor/README.md
kiendinh: It has been fixed in the repushed commit

sensors/ocf_temperature_sensor/x86/src/Makefile
kiendinh: I brought the whole Makefile from iotivity-constrained. So far it doesn't support compiling outside of the source tree. Zephyr reuses some parts of Contiki but I don't know if it is an updated version.

sensors/ocf_temperature_sensor/x86/src/Makefile
kiendinh: The -DNDEBUG seems to be used to disable assert() calls.

sensors/ocf_temperature_sensor/x86/src/main.c
kiendinh: I've updated it in the newly repushed commit.

sensors/ocf_temperature_sensor/README.md
gvancuts: How about we indicate the commits used as a ref until we have something better?

IIRC, you cannot use the 0.9 SDK from Zephyr, if true, can you add a note indicating that the SDK 0.8.2 should be used for now?

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+    (e.g. echo "connect E6:DF:CD:CE:65:14 2" >/sys/kernel/debug/bluetooth/6lowpan_control)
+
+    # ifconfig  ##You should see bt0 interface, if everything was okay
+```
+
+### Run the tests on Host:
+#### From the browser:
+
+Discovery: http://localhost:8000/api/oic/res  ## You should see the /a/temperature resource.
+Get: http://localhost:8000/api/oic/a/temperature?di=<device ID> ## You should see the temperature properties.
```
gvancuts: This breaks when viewed online (both lines appear on the same line). Can you fix that (I think an extra line may fix this or turning that into an unnumbered list would do too)

sensors/ocf_temperature_sensor/x86/src/Makefile
gvancuts: Thanks for the clarification

sensors/ocf_temperature_sensor/x86/src/Makefile
gvancuts: OK, unless @poussa or @kmaloor can enlighten us, let's keep this as is

qiuzhong:
By following the README, the Ubuntu host can discover the "OCF Temperature Node". But I didn't go to the REST API step as I failed to install iot-rest-api-server and iotivity-node on the Ubuntu host(mainly due to the network problems). I'll try it again.

gvancuts:
Thanks @qiuzhong , out of curiosity, how did you do the discovery without iotivity-node and the iot-rest-api-server?

Your comment made me realise a good point, that is in order to test as per the README, one needs to have the iot-rest-api-server snap installed. @kiendinh could you add a line that tells people this can be done on Ubuntu 16.04 using `sudo snap install iot-rest-api-server --edge`. An alternative is to install the snap that also contains the rule engine from our own demo, i.e. `sudo snap install iotivity-smarthome-demo --edge`

mjkadera:
I am a little curious about the --edge install flag and others that are available. Can you share more info?

uhofemei:
https://developer.ubuntu.com/core/get-started/developer-setup 

--edge gets the snap from the edge channel. --devmode  would give the snap unconfined access to the host device.

gvancuts:
There is a good description of the various channels [here](https://snapcraft.io/docs/reference/channels). You'll see that you have to pull those from the edge channel because of the `grade` value in the `snapcraft.yaml` files.

The notion of confinement is not directly related to the channel, you can read more about the various types of confinements [here](https://snapcraft.io/docs/reference/confinement).

qiuzhong:
@gvancuts , Here are my steps
1. Enable the Bluetooth on the Ubuntu host 
2. Scan in the bluetoothctl utility and can find a Bluetooth device named "OCF Temperature Node".
	```
	# bluetoothctl
	[bluetooth]# scan on
	[bluetooth]# scan off
	[bluetooth]# devices
	[NEW] Device F3:74:CE:45:7F:CB OCF Temperature Node
	...
	```
I think this Bluetooth device is from Arduino 101. It appears only after I build and flash the sensor app to the Arduino 101.
Of course, without the iotivity-node and iot-rest-api-server, I can't make any OCF resource discovery or get detailed OCF resource information.

Sorry for any misleading description.

gvancuts:
Thanks for the clarification @qiuzhong

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+**Note**: You can learn more information from this [wiki](https://wiki.zephyrproject.org/view/Arduino_101)
+
+* Download iotivity-constrained project source code and prepare the environment
+```
+    $ git clone https://github.com/iotivity/iotivity-constrained && cd iotivity-constrained
+    $ git submodule update --init
+    $ export IOTIVITY_CONSTRAINED_BASE=$PWD
+```
+
+In the app source folder, use the following commands to build and flash the app
```
poussa: It is not clear what app folder is. State the repo and path explicitly (sensors/ocf_temperature_sensor)

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+to Arduino 101 board:
+
+### ARC core application
+```
+    $ cd arc
+    $ make pristine && make
+    $ sudo dfu-util -a sensor_core -D outdir/arduino_101_sss/zephyr.bin
```
poussa: No need for sudo

sensors/ocf_temperature_sensor/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+    # hciconfig # You should see the device hci0
+    # rfkill unblock bluetooth
+    # hciconfig hci0 up
+    # modprobe bluetooth_6lowpan
+    # echo 1 > /sys/kernel/debug/bluetooth/6lowpan_enable
+    # bluetoothctl
+    [bluetooth]# scan on
+    [bluetooth]# scan off
+    [bluetooth]# devices # You will see the "OCF Temperature Node" with the address
```
poussa: For the hackathon, the issue here is that there is many devices with the same name. How do people know which one is them?

gvancuts: How about we add a note in the step-by-step guide to change just that part of the code before compiling and flashing the Arduino 101?

sensors/ocf_temperature_sensor/README.md
gvancuts: As a nitpick, I would prefer to use a consistent schema when naming the various folders, i.e. always use '_' (underscore) or '-' (hyphen) in between words.

poussa:
Tested & works.

gvancuts:
Thanks @poussa !

@kiendinh , can you address the last comments and I'll merge this!

sensors/ocf_temperature_sensor/README.md
kiendinh: I've changed it in the repushed patch. Let's align the folder name in the next commit. Thanks

sensors/ocf_temperature_sensor/README.md
kiendinh: It's been fixed. Thanks.

sensors/ocf_temperature_sensor/README.md
kiendinh: I've tried several ways to enable name definition in the command line like "make BLE_NAME='ABC'" by utilizing the Makefile, but then since the name is taken inside the Zephyr tree, we'll need to pull the whole ipss code here to make it happen. Now it seems changing name in the prj.conf is the simpler way.

kiendinh:
I've repushed the new patch to address most of the comments. Thanks

gvancuts:
Thanks @kiendinh , merging this.

PR #95, opened 2017-02-14
Need a way to use ADC with Mraa and MinnowBoard

odwdinc:
Created a UPM wrapper for querying ADC values for MinnowBoard MAX

PR #96, opened 2017-02-14
Ambient light sensor: Updated to use external ADC

odwdinc:
Updated to use external ADC for uses with MinnowBoard

### DIFF
--- a/ocf-servers/js-servers/ambient_light.js
+++ b/ocf-servers/js-servers/ambient_light.js
@@ -28,6 +28,9 @@ var device = require('iotivity-node'),
 var mraa = '';
 try {
     mraa = require('mraa');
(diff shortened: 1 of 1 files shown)


gvancuts:
Can you udpate the commit message to make clear what file it touches, e.g.: "ambient_light.js: update to use external ADC on MinnowBoard" (or something similar)

Could we run this new version of the script on other boards such as Edison and/or Joule?

uhofemei:
I updated the pull request commit message. I also tested the code on Edison and it works fine. :+1:

gvancuts:
Thanks for the testing @uhofemei !

I'll hold off merging this as there are changes I have requested in #95 that would affect code here (location of the wrapper library and its name).

Can the author of this PR also sign it off (`Signed-off-by` in the commit message)

PR #100, opened 2017-02-15
OCF temperature sensor: support temperature, units and range

kiendinh:
@poussa, @gvancuts
I've added the support for three properties following "oic.r.temperature" and change the default temperature unit to Fahrenheit for the event in the US. Please take a look. Thanks.


poussa:
LGTM. Did not test.

gvancuts:
Same here, have not tested this but 👍

PR #102, opened 2017-02-17
Cloud portal: add generic widget.

ichbinblau:
Add support for generic resource. 
Any unknown resources would be treated as generic sensors. 


gvancuts:
Thanks @ichbinblau , this looks good to me but haven't tested myself.

@ttzeng , could you do a sanity test on this?

gvancuts:
Hi @ichbinblau , I set out to try this yesterday before the hackathon but I couldn't get the current version of the cloud portal to behave correctly (in a Docker container), even without this PR. I ran some of the known devices (such as temperature, light, solar) in simulation mode but the actual value of the devices was not displayed in their respective widgets. I couldn't read what it said for `solar` but `temperature` was showing **NaN** instead of a value. I could see from the Docker container traces that it was getting values and we could see those from a browser tab (using `localhost/api/oic/a/temperature?di=<device ID>`). I'll file as a separate issue but I'm mentionning this here as it basically prevented me from doing any additional testing yesterday.

ttzeng:
I am sure this PR can show unknown resources in the generic widget, which presents the key-value pairs in table format, and user can edit values with the desired data type. FYI.

gvancuts:
Thanks @ttzeng , I'll merge this then based on your testing.

I would appreciate it if you and @ichbinblau could help me figure out what could have gone wrong in #105

PR #104, opened 2017-02-22
power-uart.js: Fix the ReferenceError in mraa UART mode

nagineni:
power-uart.js was failing to start when mraa UART is used instead of BLE. The issue was introduced when BLE code was added. 

```
$ node power-uart.js 
No noble module: Cannot find module 'noble'. Switching to wire connection.
SmartHome-Demo/ocf-servers/js-servers/power-uart.js:173
        uart = new mraa.Uart(dev);
                             ^

ReferenceError: dev is not defined
    at setupHardware (SmartHome-Demo/ocf-servers/js-servers/power-uart.js:173:30)

```  

### DIFF
--- a/ocf-servers/js-servers/power-uart.js
+++ b/ocf-servers/js-servers/power-uart.js
@@ -58,7 +58,7 @@ if (!noble) {
          */
         args = process.argv.slice(2);
         args.forEach(function(entry) {
-            dev = entry;
+            serialDev = entry;
         });
     }
 }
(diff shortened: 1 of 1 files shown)


nagineni:
@gvancuts @kiendinh

gvancuts:
👍 I have no tested this but this absolutely looks like the right fix and small enough that I feel comfortable merging it now.

nagineni:
Thanks :smile:

Issue #106, opened 2017-02-23
Introduce a command-line argument to start OCF JS server in simulation mode


gvancuts:
We should have a command-line parameter (e.g. `--simulation`) to start any of the OCF JS servers in simulation mode. They currently do that when some of the low-level libraries (such as `noble` or `mraa`) are not available but a switch forcing that irrespective of that would be desirable.

nagineni:
:+1:

gvancuts:
This can be closed as #116 and #112 have both been merged

PR #112, opened 2017-03-08
temperature.js: add command-line arg to force simulation mode

gvancuts:
Add a command-line argument ('--simulation') to force a simulation
mode when running the temperature OCF JS server. This decouples the
simulation mode from whether the 'mraa' module is present or not
and it can be used to more easily and predictably use that mode.


gvancuts:
First shot at addessing #106. I've only done this for the `temperature.js` server so far as I want to collect feedback from folks as to this seems like the right way of doing this. I'm especially interested in feedback from @nagineni and @poussa as I'm far from being a JS expert.

The one scenario I was not able to test here is to see if the mraa module is installed on the system. I don't have the hardware readily available but if someone could run such test ( @ttzeng , @uhofemei perhaps?), I would appreciate it.

ocf-servers/js-servers/temperature.js
```diff
@@ -33,16 +34,28 @@ var units = {
     K: 'K',
 };
 
+// Parse command-line arguments
+var args = process.argv.slice(2);
+args.forEach(function(entry) {
+    if (entry === "--simulation") {
```
nagineni: How about adding a short option (e.g `-s`) too?

ocf-servers/js-servers/temperature.js
```diff
@@ -33,16 +34,28 @@ var units = {
...
-        mraa.Aio =  require('./helper-function/mraa_MinnowBoard_ADC.js').Aio;
+if (!SimulationMode) {
+    try {
+        mraa = require('mraa');
+        if(mraa.adcSupportedBits() == 0 ){
+            mraa.Aio =  require('./helper-function/mraa_MinnowBoard_ADC.js').Aio;
+        }
+    }
+    catch (e) {
+        debuglog('No mraa module: ', e.message);
```
nagineni: I think it make sense to set `SimulationMode=true` in case of no `mraa` module found.

ocf-servers/js-servers/temperature.js
```diff
@@ -76,7 +89,7 @@ function getRange(tempUnit) {
 // This function construct the payload and returns when
 // the GET request received from the client.
 function getProperties(tempUnit) {
-    if (mraa) {
+    if (mraa && !SimulationMode) {
```
nagineni: We can remove `mraa` check here if you set `SimulationMode=true` when no `mraa` module found.

uhofemei:
👍  on the current PR. Tested in the staging environment and found it to work.

gvancuts:
Updated patch with @nagineni suggestions force-pushed.

uhofemei:
👍 tested and works.

gvancuts:
Can you tell us more about that problem @uhofemei  (debug output, etc.)?

Is it the incremental value that looks odd or is it a problem with the script not going into simulation mode?

Note that the incremental value was showing odd roundings even before the patch... not sure whether this is what you are seeing now.

uhofemei:
Output using simulation mode: 
root@motionnew:~/git/SmartHome-Demo/ocf-servers/js-servers# NODE_DEBUG=temperature node temperature.js -s
TEMPERATURE 721: Running in simulation mode
TEMPERATURE 721: Create Temperature resource.
TEMPERATURE 721: Temperature:  0.1
TEMPERATURE 721: register() resource successful

Output using sensor mode:
NODE_DEBUG=temperature node temperature.js
TEMPERATURE 754: Create Temperature resource.
TEMPERATURE 754: Temperature in Celsius:  21.66
TEMPERATURE 754: register() resource successful

gvancuts:
That's normal, try looking up the value using the SmartPhone app or the IoT REST API Server and you will see more values in the console (imcluding incremental values in case of simulation mode)

uhofemei:
It used to update every second and I was getting multiple values for each mode before the update.

gvancuts:
You most likely had something running in the system (cloud portal?). Have you tried checking the sensor (using api/oic/a/temperature?di=<di> for example?

uhofemei:
I can get updated values using the REST API. 👍

gvancuts:
Thanks @uhofemei .

@nagineni , everyone else, any other comment? If not, I'll merge this sometime tomorrow.

Issue #108, opened 2017-02-28
Buzzer not functioning when button pressed


mjkadera:
The buzzer does not make a sound when the button is pressed. The buzzer does function since it does make a noise when the motion board is restarted

gvancuts:
The following custom rule was added to the live system but did not work:
```
  "button":{
        "add":{
        },
        "update":[
            {
                "name":"value",
                "condition":"new_value === true",
                "actions":[
                    { "Type": "buzzer", "att": { "value": true }, "Event": "update" }
                ]
            },
            {
                "name":"value",
                "condition": "new_value === false",
                "actions":[
                    { "Type": "buzzer", "att": { "value": false }, "Event": "update" }
                ]
            }
        ],
        "remove":{
        }
    }
```

nagineni:
I have not tested this on the HW yet, but I quickly checked on my desktop env and the rule seems to be working fine.

$ SmartHome-Demo/ocf-servers/js-servers$ NODE_DEBUG=button  node button.js
```
BUTTON 3732: No mraa module:  Cannot find module 'mraa'
BUTTON 3732: Create button resource.
BUTTON 3732: register() resource successful
BUTTON 3732: Send the response:  true
```

$ SmartHome-Demo/ocf-servers/js-servers$ NODE_DEBUG=buzzer node buzzer.js
```
BUZZER 3716: No mraa module:  Cannot find module 'mraa'
BUZZER 3716: Create Buzzer resource.
BUZZER 3716: Update received. value:  true
BUZZER 3716: Send the response. value:  true
BUZZER 3716: Successfully notified observers.
```

$ SmartHome-Demo/gateway$ NODE_DEBUG=gateway-server node gateway-server.js
```
GATEWAY-SERVER 3681: Resource updated:{
    "id": "button",
    "value": true
}

GATEWAY-SERVER 3681: [{"Type":"button","Event":"update","att":{"id":"button","value":true}}]
GATEWAY-SERVER 3681: ----------------------------------------------
GATEWAY-SERVER 3681: Sending Update to buzzer resourceId:b675595e-bef8-47b7-a9eb-cea8b53ca0cc:/a/buzzer
GATEWAY-SERVER 3681: Resource updated:{
    "value": true,
    "id": "buzzer"
}
```

nagineni:
I checked it on the target HW and it's working fine for me. I tried with both button-toggle.js and button.js. In button.js case, you need to long press the button to change the state to true.

[...]

PR #115, opened 2017-03-08
rules.json: rule to trigger the buzzer from the button

gvancuts:
Add a rule to the Home GW rule engine (gateway/rules-engine/rules.json)
to trigger the buzzer when the button (if present) is pressed. This is,
in other words, a basic doorbell :)


gvancuts:
I haven't specifically tested this but @nagineni 's feedback (in #108) indicated that this works fine.

gateway/rules-engine/rules.json
```diff
@@ -43,4 +43,26 @@
         "remove":{
         }
     }
+    "button":{
```
nagineni: I'm sure this will through a parse error due to the missing ','.

```
},
"button":{
```

gvancuts: Good catch @nagineni , I have updated the patch but let me test it a little more then to make sure there is no other silly mistake in there ;)

gvancuts:
I've actually tested this now (using a simulated buzzer and button) and the output (using NODE_DEBUG=buzzer and NODE_DEBUG=button) shows that the rule is in effect. This should be good to go in.

nagineni:
LGTM. @gvancuts thanks for testing. :smile:

PR #117, opened 2017-03-10
temperature.js: cosmetic change for the SimulationMode variable

gvancuts:
Cosmetic change from 'SimulationMode' to 'simulationMode' as
this appears to be a more widely used convention, at least in
this repository.


uhofemei:
👍

PR #126, opened 2017-04-06
Provide Gateway and SmartDevices in Docker containers

### LINKED ISSUE #123: Containerize the Gateway and Smart Devices (separately)
The plan is to have two container services:
* Gateway that will include IoT REST API, IoTivity and rules engine
* Sensor container that will simulate the available sensors

When we scale that gateways, we deploy the gateway and the sensor container. The admin user will login to the admin portal to add each gateway.

One concern is the message queue will not scale. Theresa is exploring if the message queue and workers can be scaled like other applications. We will need 20 gateways for the OpenStack Summit

We will also need to test the performance of the MySQL server database.

### LINKED ISSUE #127: Enhance `smarthome-sensors` Docker container: selectively start some of the OCF servers
We should enhance the `smarthome-sensors` Docker container by providing a mechanism to allow to selectively choose which OCF servers to start (in simulation mode) inside any container running the image. The exact mechanism on how to do that is to be defined, it could be a shared file listing the OCF servers to start, or a list of them passed as arguments, etc.

gvancuts:
This introduces 2 new Docker containers that provide the following:
* Gateway in a container. Offers the following services:
   * IoT REST API Server
   * Gateway rule engine
   The Dockerfile describing it is located under gateway/
* SmartDevice in a container: it runs all OCF (JavaScript) servers
  in simulated mode. The Dockerfile describing it is located under
  ocf-servers/

Fix #123 
Fix #127 


gvancuts:
Note that the `smarthome-[sensors|gateway]` will only be available from Docker Hub once this PR has been merged.

ichbinblau:
Hi, @gvancuts. I have created a  PR https://github.com/gvancuts/SmartHome-Demo/pull/2 for the issue #127 . PTAL

ichbinblau:
Additionally, I think we could remove the folder `smarthome-web-portal/tools/test_harness/` which was created for emulating gateway. It is a duplicate.

qiuzhong:
Test results for this PR:
* Running the containers within the docker0 bridge network
  Both containers can work well. It can also work with the smart-home-portal cloud app.

* Running the containers directly on the host network.
  <code>smarthome-gateway</code> REST API server cannot work well.
  On host, sending requests to <code>http://localhost:8000/api/oic/d</code> always gets <code>[]</code>. 
  For <code>http://localhost:8000/api/oic/p</code> and <code>http://localhost:8000/api/oic/res</code>, you can also get <code>[]</code> occasionally.

  Not sure if it's related to the docker network configuration.

From both options, port <strong>8000</strong> and <strong>8080</strong> are not used for container <code>smarthome-sensors</code>. Is <code>EXPOSE 8000 8080</code> really necessary for that container?

gvancuts:
Thank you for your testing @qiuzhong !

Regarding the `EXPOSE 8000 8080` line in `Dockerfile`, you're right, this is not needed, I'll remove it. (I actually believe that it's not needed ever and that the command-line argument is sufficient but let's keep it in for the smarthome-gateway `Dockerfile` anyway.

I'm not sure what's wrong in your test when using the host network. It seems like the `IoT REST API Server` is running correctly but not seeing any OCF server. What does http://localhost:8000/api/system return for example? Are you sure the `smarthome-sensors` was correctly started using the same host network option?

gvancuts:
Actually, you should still see more than just `[]` regardless of whether there are any OCF servers around. Could you possibly have any service on the host already using that port?

gvancuts:
New version (that also includes @ichbinblau latest feature) force-pushed.

gvancuts:
Merging

gvancuts:
Now that this PR has been merged, you can grab these Docker containers directly from Docker Hub, i.e.:
```
$ docker pull smarthome2cloud/smarthome-sensors
$ docker pull smarthome2cloud/smarthome-gateway
$ docker pull smarthome2cloud/smarthome-cloud
```

uhofemei:
I can't access the REST API after starting the smarthome-gateway container. 

```demouser@gateway:~/SmartHome-Demo$ docker run smarthome2cloud/smarthome-gateway
This container IP address is: 172.17.0.2
Press [CTRL+C] to stop..
API server started on 2017-04-26T20:19:55.689Z [port: 8000, https: false]```

PR #142, opened 2017-05-23
snap: expose 'node' command for running user scripts

ttzeng:
This commit exposes the built-in `node` command for running user given scripts such as the OCF servers provided in this repository, the required node.js module dependencies are pre-installed.

Signed-off-by: Tonny Tzeng <tonny.tzeng@intel.com>


snap/README.md
```diff
@@ -75,6 +75,8 @@ bluez:service       bluez:client,iotivity-smarthome-demo:bluez
 In the SSH session with the Intel Joule development kit installed this snap, reference the following list for the defined commands and their functions:
 * **run-gateway-server**  
   This command launches the Smart Home gateway server which supports 3D webGL UI at port 8080, and run the rule engine that handles the interactions between discovered OCF servers.
+* **node** &lt;user script&gt;
```
gvancuts: Tested this and it worked for me! One suggestion would be to add a practical example to the readme. E.g.
```
iotivity-smarthome-demo.node /patch/to/SmartHome-Demo/ocf-servers/js-servers/gas -s
```

ttzeng: @gvancuts thanks for providing feedbacks. The amended PR has been added an example as you wish.

Regarding the warning message of the `network-bind` plug, what's your testbed? were servers can be discovered after getting this message? the `network-bind` interface should be auto-connect during snap install...

gvancuts: Yes, the servers were discovered correctly. It does not seem to affect the functionality of the snap. I suggest that we do not hold up merging this because of that issue. If it is persistent and/or affect the functionality, we'll have to look into it.

PR #144, opened 2017-05-26
Update instructions: sudo, 01org path, DH pull fixes

mmccool:
Some updates to instructions to fix the following issues:
- Add sudo to all commands.  Making docker run without sudo is a security hole, and this was also noted.  If people set things up to run without it, they can just cut-and-paste commands without the sudo.
- Add 01org to git clone path (it was missing and therefore, broken)
- Add alternative commands for running from images pulled from Docker Hub.   The previous commands only worked for locally built images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,16 +31,16 @@ You will find more information about these containers in the `README.md` files r
 
 * Install Docker on your host OS following the official documentation at [Docker Platform Installation](https://www.docker.com/products/overview#/install_the_platform).
 
-*Note:* you need to be `root` (`sudo`) by default on most host OSs. The commands listed below assumed you have configured your system so you can run `docker` as a normal user.
+*Note:* You need to be `root` (or use `sudo`) by default to run `docker` on most host OSs. Configuring `docker` to run as an ordinary user is possible (one way is to add that user to the `docker` group) but currently compromises security.  If you do this however you can just drop the `sudo` prefixes in the examples below.
 
 * Clone the `SmartHome-Demo` repository
 ```
-$ git clone https://github.com/SmartHome-Demo
+$ git clone https://github.com/01org/SmartHome-Demo
 $ cd SmartHome-Demo
 ```
 
 ### Getting (any of) the Smart Home containers
-There are two options available, build it locally or grab it from [Docker Hub](https://hub.docker.com).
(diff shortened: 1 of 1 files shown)


mmccool:
Also... the instructions don't show running the cloud service.

README.md
```diff
@@ -53,33 +53,46 @@ It will take a little while depending on your host machine as it compiles [IoTiv
...
 ```
 
 #### Running the Docker containers
 There are a few ways you can run the containers, one critical aspect is to get the networking part right. The `smarthome-sensors` and `smarthome-gateway` containers need to be on the same subnet in order for the OCF servers to be discovered by the gateway. This is the case by default if you run both containers on the same host machine (by default, both containers will be running in the `docker0` bridge network). If you intend to run those on different physical machines (that are on the same subnet), the easiest way is to use the `--network host` parameter. If you are interested in understanding more about networking and Docker, please visit the [Docker container networking](https://docs.docker.com/engine/userguide/networking/) page. 
 
 
 **Option 1:** Running the containers within the `docker0` bridge network
-When running both `smarthome-sensors` and 'smarthome-gateway` containers on the bridge network, we recommend publishing port 8000 to a local port (e.g. 8001). Example:
+When running both `smarthome-sensors` and 'smarthome-gateway` containers on the bridge network, we recommend publishing port 8000 to a local port (e.g. 8001).
```
gvancuts: There is an apostrophe just before `smarthome-gateway` that should be replaced by a backtick to get the right formatting.

gvancuts:
> Also... the instructions don't show running the cloud service.

Yes, we have some information in this README.md file: `smarthome-web-portal/tools/docker/README.md`. But we need to:
1. Update those instructions (they're slightly out of date)
2. Refer to that README.md more explicitly

Thanks,
Geoffroy

gvancuts:
Created #145 to make sure we fix and improve the instructions related to running the cloud portal.

mmccool:
Fixed issues noted.   I'll take a stab at getting the cloud service running with the instructions provided (and, assuming I can figure it out, will suggest updates to them while I'm at it...)

gvancuts:
Thanks @mmccool , I'll merge this now.

gvancuts:
@mmccool let us know if you have any problems getting the cloud running... and any PR to improve the documentation would certainly be greatly appreciated ;-)

PR #148, opened 2017-06-01
gateway: enable cli arguments for the IoT REST API Server

### LINKED ISSUE #147: Add a parameter to the smarthome-gateway container to switch between HTTP and HTTPS
We should add support for a parameter (`-s` recommended) so that the smarthome-gateway container provides support for HTTPS (instead of plain HTTP).

gvancuts:
* The IoT REST API Server Command-line arguments are enabled in Docker
  and transparently passed through. This allows to tweak its behaviour
  by changing the port and/or enabling https
* The README.md was updated to describe a couple of methods for deploying
  the Home Gateway function

Fix #147 

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -31,6 +31,9 @@ WORKDIR /opt/SmartHome-Demo/gateway/
 
 RUN npm install
 
+# Install the certificates needed when using HTTPS
+RUN /bin/bash /opt/SmartHome-Demo/gateway/node_modules/iot-rest-api-server/config/generate-key-and-cert.sh
+
 # Forward port
 EXPOSE 8000 8080
 
@@ -39,5 +42,5 @@ ENV http_proxy ""
 ENV https_proxy ""
 
 # Start Home Gateway
-CMD ["/opt/SmartHome-Demo/gateway/start-gateway-in-docker.sh"]
+ENTRYPOINT ["/opt/SmartHome-Demo/gateway/start-gateway-in-docker.sh"]
 
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -31,8 +31,7 @@ WORKDIR /opt/SmartHome-Demo/gateway/
 
 RUN npm install
 
-# Install the certificates needed when using HTTPS
-RUN /bin/bash /opt/SmartHome-Demo/gateway/node_modules/iot-rest-api-server/config/generate-key-and-cert.sh
+VOLUME ["/opt/security-files/"]
 
 # Forward port
 EXPOSE 8000 8080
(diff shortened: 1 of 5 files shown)


gvancuts:
@poussa @nagineni @uhofemei @ttzeng @ichbinblau @mjkadera please take a look

gateway/Dockerfile
```diff
@@ -31,6 +31,9 @@ WORKDIR /opt/SmartHome-Demo/gateway/
 
 RUN npm install
 
+# Install the certificates needed when using HTTPS
+RUN /bin/bash /opt/SmartHome-Demo/gateway/node_modules/iot-rest-api-server/config/generate-key-and-cert.sh
```
ichbinblau: Hi, @gvancuts.  I have a minor finding. The script `generate-key-and-cert.sh` takes one parameter called FQDN. The FQDN is supposed to be the same as the service external facing host name (www.xxx.com etc) or simply the external IP address.  By default, it uses localhost which is 127.0.0.1 which is not right. As the Cloud portal does not validate the SSL certificate, it is not a problem for the time being.  If we are going to make the site secure in future, it will be a potential issue.

gvancuts: Thanks @ichbinblau , this is good to know. We'll have to figure out a way of passing those certificates to a Docker container then (and remove that line) as we won't know what those are at build time. I'm assuming we will have to use the [data volume](https://docs.docker.com/engine/tutorials/dockervolumes/) mechanism for that (just like what you use for selectively starting OCF servers). It shouldn't be too hard, I'll take a look at it.

gvancuts:
New, enhanced version has been pushed. Let me know if you like the approach @ichbinblau (and all others). If so, I really ought to document it too before merging this ;)

gateway/start-gateway-in-docker.sh
```diff
@@ -11,17 +11,57 @@ export NODE_PATH=/opt/SmartHome-Demo/gateway/node_modules/
...
+		if [ "$arg" = "-s" ]
+		then
+			echo
+			echo "*** WARNING ***"
+			echo
+			echo "No certificate or private key was found although you are requesting to use 'https'. We highly recommended that you provide a valid certificate and private key."
+			echo
+			echo "We will proceed using a certificate and private key that was generated for testing purposes only but be aware that the certificate is self-signed and browsers do not recognise it so you will get warnings."
+			echo
+			echo "We highly recommende that you get a proper certificate from a known certificate authority and its corresponding private key if you want to use 'https'"
```
ichbinblau: A minor typo: recommende -> recommend.

ichbinblau:
@gvancuts, I like the enhancement, the warning on startup in particular. I have verified this PR with https enabled/disabled and both work.

gvancuts:
Nice catch @ichbinblau ! I have fixed the typo and have also now added some proper documentation in the README.md files to explain what this is and how to use it. I am planning to merge this next Tuesday if there is no concerns from anyone.

gvancuts:
I haven't heard any concerns or objections, merging this.

PR #160, opened 2017-06-22
Update and move all systemd files provided as a reference

gvancuts:
The content of the 'ostro-config' folder is very incomplete and
obsolete at this stage. Since we do not use this anywhere, let's
remove it.


ttzeng:
I can confirm I never use stuffs in the `ostro-config` folder, as I used to disable the iptables completely, and launch the scripts directly from the `gateway/ocf-servers` folder.

kiendinh:
I think it has some value as sample code for systems that use systemd, and have firewall with iptables. Security is necessary and opening necessary ports just enough for the services like IoTivity is always a good practice. The name probably should be changed, though.

ttzeng:
@kiendinh if we want to keep these files, we may need to revise the script paths, or document somewhere to copy those scripts from the `ocf-servers/js-server` folder to `/home/root`, otherwise these systemd service files launch the servers from the wrong location.

ttzeng:
In addition, there is only config file for the gateway, we'll still need additional configure files to open the coap ports. Besides, random user ports will be required to be enabled for the OCF client.

kiendinh:
Fair enough. I hoped to keep the value with the least work to do. But deleting it sounds easier. So no objection.

nagineni:
Those scripts were added long ago for convenience purposes and we never spend time on maintaining them. How about re-naming the directory and fixing those services instead of fully removing? It makes lot easier to start ocf servers automatically on boot when you have systemd services in place.

nagineni:
@uhofemei

gvancuts:
You've kind of convinced me that there is value in putting a little more efforts into these to not drop them completely. I had done (a while ago) some work on that so let me see if I can still find that branch and re-use some of it...

uhofemei:
@gvancuts if you can't find your branch I would be happy to submit a PR containing the systemd service files.

gvancuts:
I found it... it's here: https://github.com/gvancuts/SmartHome-Demo/tree/enhanced-systemd-files

But it still really needs quite a bit of work (I thought I had done a bit more than that :-))

Feel free to take it from there or start from scratch if you want to tackle this!

gvancuts:
I have finished up what was in my branch and this PR has been completely changed.

PTAL

gateway/README.md
```diff
@@ -55,29 +55,71 @@ For more details, please take a look at the [Dockerfile](./Dockerfile).
...
+     $ /usr/bin/node node_modules/iot-rest-api-server/index.js &
+  ```
+
+2. Automating things with [systemd]
+
+For those Linux operating systems that use `systemd`, we provide `systemd` service files in the [`systemd-files`](./systemd-files) folder. Follow these few steps below to use those:
+```
+$ sudo cp -r /opt/SmartHome-Demo/gateway/systemd-files/* /usr/lib/systemd/system/
+$ sudo systemctl daemon-reload
+$ sudo systemctl start smarthome-gateway
```
gvancuts: What OS have you tested this with?

gateway/README.md
```diff
@@ -55,29 +55,71 @@ For more details, please take a look at the [Dockerfile](./Dockerfile).
...
+  ```
+     $ cd /opt/SmartHome-Demo
+     $ /usr/bin/node node_modules/iot-rest-api-server/index.js &
+  ```
+
+2. Automating things with [systemd]
+
+For those Linux operating systems that use `systemd`, we provide `systemd` service files in the [`systemd-files`](./systemd-files) folder. Follow these few steps below to use those:
+```
+$ sudo cp -r /opt/SmartHome-Demo/gateway/systemd-files/* /usr/lib/systemd/system/
```
gvancuts: Why can you not put those in /usr/... ?

uhofemei: There's no system folder in /usr/lib/systemd and I never put them there.

gateway/README.md
uhofemei: Ubuntu 16.04

gateway/README.md
gvancuts: Fedora has both of them, I'll update the code to use /lib/... since that one is common between the two distros.

gateway/README.md
gvancuts: Thanks. ip(6)tables is under /usr/sbin on Fedora... I'll switch to your recommendation (which works for Ubuntu) and add a note that the path should be updated on Fedora and possibly other distros.

gvancuts:
Updated version force-pushed, PTAL.

uhofemei:
LGTM 👍

uhofemei:
Could /usr/sbin/ip(6)tables on Fedora be a symlink pointing to /sbin?

gvancuts:
> Could /usr/sbin/ip(6)tables on Fedora be a symlink pointing to /sbin?

It turns out Fedora has ip(6)tables under both /sbin and /usr/sbin. All of these are symlinks to a binary called xtables-multi found in, respectively, /sbin and /usr/sbin (these binaries are identical).

gvancuts:
If no other comments from anyone, I will merge this in about 12 hours.

nagineni:
:+1:

PR #167, opened 2017-06-30
Clean up unnecessary dependencies.

ichbinblau:
1. Remove unnecessary tools and dependencies
2. Remove python dependency's dependencies.  
    For instance, Werkzeug and Jinja2 are Flask's dependency. They will be installed together with Flask. Hence, remove them from the requirements.txt
3. Replace mysql server with mariadb server
    The db initialization and password reset actions were moved to run-time phrase. The `start-service.sh` was updated to control the startup sequence.


smarthome-web-portal/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN echo $http_proxy
 RUN apt-get clean && apt-get update
 
 # install python2 and common tools
-RUN apt-get -y install vim wget netcat python
+RUN apt-get -y install netcat python
```
gvancuts: Is netcat a real dependency (or more of a debug tool that could also be removed)?

qiuzhong:
I tried and it worked for me so far. 
However, I had to use `--no-cache` when building the docker image. Otherwise, it would stop at  
`E: Unable to locate package netcat` and `E: Unable to locate package python` errors.

I believe `netcat` is not necessary as it's a network utility. 
Finally, the `sources.list` is not used by the `Dockerfile` anymore. So I think it's not necessary either.

gvancuts:
> I tried and it worked for me so far.
Thanks for checking this!

> However, I had to use --no-cache when building the docker image. Otherwise, it would stop at
> E: Unable to locate package netcat and E: Unable to locate package python errors.
> I believe netcat is not necessary as it's a network utility.
Interesting... I'd like @ichbinblau to remove netcat if it's not a real dependency but it sounds like in  your case, you'd still hit the same issue with python. It does not seem fatal though and assuming Docker Hub is able to build these images just fine, I don't believe we should consider this blocking, do you agree?

> Finally, the sources.list is not used by the Dockerfile anymore. So I think it's not necessary either.
That's true, they are there as an example of how one could use mirror sites instead of the main repos. There are lines in Dockerfile that are commented out that refer to those files.

qiuzhong:
@gvancuts ,

> I don't believe we should consider this blocking, do you agree?

Yes, you're right. I agree with you. I had to use `--no-cache` because I was building the docker image on a Ubuntu system that I used to build the portal image before. I guess the `RUN apt-get update` instruction failed for some reason(Maybe cache). So it complained `E: Unable to locate package xxx`. 

I just tried on another Ubuntu system without building the portal image before and such errors were not found.

ichbinblau:
@qiuzhong @gvancuts Netcat was used to check the open ports in `start-service.sh`. But it is not necessary,  I've replaced it with `netstat` from package net-tools which is in the white list.

uhofemei:
I don't think netstat is installed on most Linux distros by default. Could you replace netstat with iptables? I believe iptables is part of every distro by default.

gvancuts:
> @qiuzhong @gvancuts Netcat was used to check the open ports in start-service.sh. But it is not necessary, I've replaced it with netstat from package net-tools which is in the white list.
:+1:

ichbinblau:
@uhofemei yes, netcat is not a native tool for minimal installation. I have replaced it with net-tools which is in the open source PDT whitelist  (https://whitelisttool.amr.corp.intel.com/)

gvancuts:
@uhofemei , can `iptables` do what @ichbinblau needs to do?

Can you explain in more details what the issue is (with regards to using a tool that's not pre-installed in most distributions)?

uhofemei:
I've been talking to Theresa about this last night. The goal is to reduce the container size footprint by not having to rely on the net-tools package to check if a port is open and a service is running. This can be done in many other ways, such as iptables (only check if the port is open) or by using tools such as fuser <port>/protocol, i.e. fuser 80/tcp, which then returns the PID for the web server that is running. All of this goes down to the fact that supervisor in the docker container doesn't do a good job starting services that are required sequentially. Instead it attempts to parallelize things, which often fails. Theresa said she would be looking at the fuser solution, which means we could remove net-tools from the BOM. I also reached out to the security team today and will submit the BOM components, which have not been entered into the white listing tool.

ichbinblau:
@uhofemei , I've removed the dependency on the net-tools package and replaced MySQL with MariaDB. Instead of checking running ports, I used their management tools `mysqladmin` and `rabbitmqctl` to check service status respectively. In that case, extra tools are not necessary anymore.

uhofemei:
LGTM 👍

gvancuts:
> LGTM 👍

Thanks for the thumbs up @uhofemei. Can I assume you have tested this as well?

uhofemei:
Yes.

gvancuts:
> Yes.

Thanks! Merging...

PR #168, opened 2017-07-31
Create stable branch and remove unstable parts.

uhofemei:
Signed-off-by: Ulf Hofemeier <ulf.hofemeier@intel.com>

PR #171, opened 2017-08-04
Create stable branch and remove unstable parts.

uhofemei:
NOTE: In the stable branch code known to run unstable has been
removed from the demo. This is currently the case with sensor sample
code for Zephyr and Zephyr.js using iotivity-constrained.

Signed-off-by: Ulf Hofemeier <ulf.hofemeier@intel.com>


gvancuts:
What about the `ocf_temperature_sensor` code? You had removed it in #168 initially. Is it deemed stable enough to be kept in the stable branch? (I know we've had much less stability problems with OCF servers having a single resource but I want to make sure you're keeping it intentionally)

uhofemei:
That one is using the board internal BMI 160 chip. The temperature reported is the Quark internal temperature, which tends to be quite a bit higher than the environmental temperature. IMHO this distinction has been confusing to people in the past. If it were up to me, I would not report the value.

gvancuts:
> That one is using the board internal BMI 160 chip. The temperature reported is the Quark internal temperature, which tends to be quite a bit higher than the environmental temperature.

That's not correct. The code in `ocf_temperature_sensor` uses a Grove temperature sensor hooked up to the board.

Besides that point, the key question is whether this is stable or not. This is what should determined whether it is included in the `stable` branch or not. You had removed it in #168 but are keeping it in this PR... what has changed?

uhofemei:
Nothing, I just missed it.

PR #174, opened 2017-08-14
Update OCF servers to the latest JS spec and iotivity v1.3.0

nagineni:
This commit updates all OCF servers to the latest JS spec and
iotivity v1.3.0. It also includes the scripts that are needed to 
create secure configurations.

### DIFF
--- a/gateway/package.json
+++ b/gateway/package.json
@@ -15,7 +15,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "iot-rest-api-server": "0.4",
+    "iot-rest-api-server": "0.6.0",
     "express": "*",
     "websocket": "*",
     "http": "*"
--- a/ocf-servers/README.md
+++ b/ocf-servers/README.md
@@ -2,8 +2,11 @@
 ## JavaScript OCF servers
 A number of JavaScript-based OCF servers are provided in the [`js-servers`](./js-servers/) and [`zjs-servers`](./zjs-servers/) folders for `node.js` and [`zephyr.js`](https://github.com/01org/zephyr.js) respectively. The `node.js` dependencies (that can be installed using `npm`) are:
 * [IoTivity-node](https://www.npmjs.com/package/iotivity-node)
+* [lodash.mergewith](https://www.npmjs.com/package/lodash.mergewith)
+* [lodash.assignin](https://www.npmjs.com/package/lodash.assignin)
 * [mraa](https://www.npmjs.com/package/mraa)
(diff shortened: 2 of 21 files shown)


gvancuts:
Thanks @nagineni ! I will test this asap (using the Docker containers - now available in `smarthome2cloud/smarthome-[sensors|gateway|cloud]:container-staging`).

One of the three commits is missing the `Signed-off-by` (https://github.com/01org/SmartHome-Demo/pull/174/commits/ddece196fe425e44e2aa2d50ca3a098704a176e3), can you add it?

nagineni:
Thanks @gvancuts. I updated the commit message now.

gvancuts:
Thanks @nagineni ! I will defer to @uhofemei and @ttzeng to determine when to merge this. Am I assuming correctly that the Zephyr-based devices have not been tested with this new version of IoTivity? I would also like to hear from @ttzeng on the readiness of the Android companion app and the JukeBox based on Android Things.

gvancuts:
Thanks @nagineni ! I will defer to @uhofemei and @ttzeng to determine when to merge this. Am I assuming correctly that the Zephyr-based devices have not been tested with this new version of IoTivity? I would also like to hear from @ttzeng on the readiness of the Android companion app and the JukeBox based on Android Things.

ttzeng:
@gvancuts this PR [patches](https://github.com/01org/SmartHome-Demo/blob/2ee59c85ee6eeea917e534efa5267ab93e173ecf/ocf-servers/js-servers/config/json-to-cbor.js#L21) the servers to have an ACE for `coap://` connections, it will be transparent to existing clients, including webUI, iot-rest-api-server, and the Android companion app, while the secured mode is ON, so the current companion app can access to the server as is. Though I only validated with the actual LED server, I believe other servers should be working fine. Having said that, this PR is safe for merging if we want to upgrade to IoTivity 1.3.0 as the 1st step, though we do not take any security features for now.

PS: the servers with [this PR](https://github.com/01org/SmartHome-Demo/blob/2ee59c85ee6eeea917e534efa5267ab93e173ecf/ocf-servers/js-servers/config/security.boilerplate.json#L40) will be in RFOTM state, the coming companion app will be able to move them to RFNOP state. And, the iotivity-node has [an issue](https://github.com/otcshare/iotivity-node/issues/145) while registering secured resources, once we fix this upstream bug, the companion app can be used for pairing the devices and make use of `coaps://` connections.

gvancuts:
Thanks @ttzeng and @nagineni , I'll merge this now!

PR #177, opened 2017-09-01
Alter 'bmi160' to 'a' in temperature.js

xiaoyu3x:
Type| Pass | Fail |Block
---| ---| ---| ---
manual |1 |0|0


gvancuts:
Hi @xiaoyu3x , why should we change this? Is it fixing a particular issue?

xiaoyu3x:
@gvancuts , you are right. I have a smarthome web portal case that shows sensor on docker cloud app(http://localhost:3030), but sensor can be obtained only when resource path format likes '/a/***'.
I'd like to change '/bmi160/temperature' to '/a/temperature' for smarthome web portal test.

gvancuts:
Hi @xiaoyu3x , if I recall correctly, we decided to use that prefix to make it clear it is not an external temp sensor. What it reports is in fact the internal temperature of the processor (and as such is only very loosely correlated with the real external temperature).

The cloud portal should still report that value although it will refer to that as a "Generic Sensor" instead of recognizing it is a temperature sensor. Are you seeing it?

@ttzeng , what is your opinion on this PR?

ttzeng:
@xiaoyu3x I have no issue to rename the path of the temperature server in short term, but it makes more sense to me to accept any URL in your web app as long as it compliants with OCF spec. You may reference the section 6.2 of the OCF core spec, "`the / (slash) separated segments are treated as a single string ... and not parsed as a hierarchy`" ... "`Once a path is generated, a Client accessing the resource or recipient of the URI should use that path as an opaque string and should not parse to infer a structure...`", that would make your web app more OCF compliant.

xiaoyu3x:
@gvancuts, Sorry for my late answer. To
> > The cloud portal should still report that value although it will refer to that as a "Generic Sensor" instead of recognizing it is a temperature sensor.
I test this issue again.  Sensor still isn't showed on the cloud portal, whether it is  "Generic Sensor"  or "Other Sensor".

gvancuts:
I suspect the issue @xiaoyu3x is facing is due to this conditional statement in the `smarthome-web-portal`: https://github.com/01org/SmartHome-Demo/blob/master/smarthome-web-portal/RestClient/Resource.py#L70

@ichbinblau , based on what @ttzeng said above, shouldn't we remove the last condition altogether (and allow any kind of `href` to be inserted into the `sensors` list?

ichbinblau:
@gvancuts Agree. I will fix it after the event next week.

gvancuts:
Thanks! 👍

gvancuts:
@xiaoyu3x , are you OK if we discard this PR and instead get this fixed (later) in the cloud portal?

gvancuts:
FWIW, I ran a quick test after removing the check on the `href` property but this is not right. Not all sensors are showing up correctly and those that do come up late and disappear. I suspect that this is too simple a fix and that we end up with a bunch of stuff in the sensor list that we don't want there (things like /oic/sec/cred, /oic/sec/acl2, etc.). We would need a more sophisticated method (maybe filtering based on the the `rt` property?)... to be continued.

ichbinblau:
Yes, it is also my concern. That's why I did not fix straightaway yesterday cos a pattern has to been found to filter the resource href.

ttzeng:
The prefix '/oic/' is reserved and shall not be used for application resources, you could try to ignore those reserved resources. BTW, it seems you have secured devices so you got /oic/sec/cred, /oic/sec/acl2 resources.

gvancuts:
Thanks for the suggestion @ttzeng , this seems like an excellent way of filtering out things we do not want. I have created a PR (see #180 ) in an attempt to use that and I'm also filtering out those that have an `href` that starts with `/introspection`.

Now, this seems to make the Cloud Portal correctly add the `/bmi160/temperature` sensor to the list but it's not yet being displayed... I don't think it's a problem in the cloud portal code anymore though as I'm unable to retrieve the data from that sensor when using the `iot-rest-api-server` directly... I need to investigate more but I'm posting the PR anyway so @ichbinblau and others can already take a look and provide feedback.

gvancuts:
Closing this and tracking the issue in #181

gvancuts:
> BTW, it seems you have secured devices so you got /oic/sec/cred, /oic/sec/acl2 resources.
That's correct. I do all this testing and fiddling on top of master... so things are all based on iot-rest-api-server 0.6.0 and iotivity-node 1.3.0-1.

ttzeng:
> Now, this seems to make the Cloud Portal correctly add the /bmi160/temperature sensor to the list but it's not yet being displayed... I don't think it's a problem in the cloud portal code anymore though as I'm unable to retrieve the data from that sensor when using the iot-rest-api-server directly

This is an [upstream bug](https://github.com/01org/zephyr.js/issues/1481) which prevent ZJS servers from being discovered.

wanghongjuan:
Hi @gvancuts @ttzeng AFAIK, the coming new release v0.4 of zephyr.js will not support OCF secret spec. Shall we enable this sensor on master and use `iot-rest-api-server` v0.5.0 to test it at first? Once the zephyr.js support OCF secret spec and align with iotivity-node 1.3.0-1, may be need to update all zjs-servers scripts.

gvancuts:
What I'm hitting is a different issue than https://github.com/01org/zephyr.js/issues/1481 as I am not using ZJS at all in my experiments. I'm using the `ocf-servers/js-servers/temperature.js` and a modified version of it, take a look at https://github.com/01org/iot-rest-api-server/issues/74 for more details on exactly how I test this. The `iot-rest-api-server` version I'm using is 0.6.0 (and I'm using `iotivity-node` 1.3.0-1)

PR #178, opened 2017-09-04
OCF JS servers: Print device ID when OCF server started

nagineni:
Print device ID in debug messages when OCF server started. The is often helpful to identify the device after resource discovery from client.


nagineni:
@uhofemei

uhofemei:
LGTM. I've tested change with the Edison OCF server code by manually merging the patch. Works well. 👍

PR #180, opened 2017-09-06
smarthome-web-portal: enable displaying of *all* sensors

### LINKED ISSUE #181: Only devices whose 'href' starts with "/a/" or "/brillo" are displayed in the portal
The cloud portal currently only display devices/sensors (OCF servers) whose `href` parameter starts with either `/a/` or `/brillo`. It is however perfectly valid to use a different prefix. We have an example of such device in our repo in `ocf-servers/zjs-servers/temperature.js` (`href` is `/bmi160/temperature`).

gvancuts:
This patch will make the Cloud Portal display all OCF servers
discovered even when their path ('href') does not start with "/a/".
We are still filtering out from the list of devices discovered
those whose 'href' starts with "/oic/" or "/introspection" as
they are not representing OCF servers (see also the OCF specs
for more details)

Fix #181 

### DIFF
--- a/smarthome-web-portal/RestClient/Resource.py
+++ b/smarthome-web-portal/RestClient/Resource.py
@@ -66,8 +66,8 @@ class Resource(object):
                         elif bm == 3 and not secure:
                             observable = True
 
-                    if rt and not rt.startswith("oic.wk") and href \
-                            and (href.startswith("/a/") or href.startswith("/brillo/")):
+                    if rt and not href.startswith("/oic/") and \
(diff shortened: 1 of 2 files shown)


ichbinblau:
Hi, @gvancuts . I have reviewed and tested the changes. The filter works perfectly.

gvancuts:
Thank you @ichbinblau . I'll merge this now. I'm still puzzled as to what is going on in https://github.com/01org/iot-rest-api-server/issues/74 but I think these are not related.

